PR #34, opened 2023-07-21
feat: add request config

liulinboyi:
Add request config
eg: For `Proxy` request:

examples\calendar\src\main.ts
```ts
import fs from "fs";
import path from "path";
import dotenv from "dotenv";
import { createLanguageModel, createJsonTranslator, processRequests } from "typechat";
import { CalendarActions } from './calendarActionsSchema';
import { SocksProxyAgent } from 'socks-proxy-agent' // 7.0.0

// TODO: use local .env file.
dotenv.config({ path: path.join(__dirname, "../../../.env") });

const PROXY_HOST = "127.0.0.1";
const PROXY_PORT = "1080";
const model = createLanguageModel(process.env, 
    {
        httpsAgent: new SocksProxyAgent({
            hostname: PROXY_HOST,
            // userId: 'xxx@gmail.com',
            // password: 'xxxxx'
            host: PROXY_HOST,
            port: PROXY_PORT,
            rejectUnauthorized: false,
        } as any)
    }
);
const schema = fs.readFileSync(path.join(__dirname, "calendarActionsSchema.ts"), "utf8");
const translator = createJsonTranslator<CalendarActions>(model, schema, "CalendarActions");
translator.validator.stripNulls = true;

// Process requests interactively or from the input file specified on the command line
processRequests("📅> ", process.argv[2], async (request) => {
    const response = await translator.translate(request);
    if (!response.success) {
        console.log(response.message);
        return;
    }
    const calendarActions = response.data;
    console.log(JSON.stringify(calendarActions, undefined, 2));
    if (calendarActions.actions.some(item => item.actionType === "unknown")) {
        console.log("I didn't understand the following:");
        for (const action of calendarActions.actions) {
            if (action.actionType === "unknown") console.log(action.text);
        }
        return;
    }
});

```

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -38,17 +38,17 @@ export interface TypeChatLanguageModel {
  * If none of these key variables are defined, an exception is thrown.
  * @returns An instance of `TypeChatLanguageModel`.
  */
-export function createLanguageModel(env: Record<string, string | undefined>): TypeChatLanguageModel {
+export function createLanguageModel(env: Record<string, string | undefined>, requestConfig?: Record<string | symbol, any>): TypeChatLanguageModel {
     if (env.OPENAI_API_KEY) {
         const apiKey = env.OPENAI_API_KEY ?? missingEnvironmentVariable("OPENAI_API_KEY");
         const model = env.OPENAI_MODEL ?? missingEnvironmentVariable("OPENAI_MODEL");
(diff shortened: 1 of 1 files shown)


robgruen:
I think we would want a stricter approach to providing request configuration or maybe even a hook as an alternate approach.  Passing through credentials and other unknowns is something we probably want to avoid.

PR #36, opened 2023-07-21
Add play playlist functionality to music example

DugganB:
Added the functionality to play a playlist in the music example. This will populate the queue in the Spotify player so you can go forward/back songs.

Before, the first song in the playlist was being played. This does not populate the queue in the Spotify player, so you can not continue listening to the playlist or go to the next song.

Changes:
- Added `getPlaylist` to `chatifyActionsSchema.ts`
- Added `Playlist` type to `chatifyActionsSchema.ts`
- Updated `play` to take in `item: TrackList | Playlist`
- Added `contextUri` param to `play()` in `music/src/endpoints.ts`

image

Note:

The `Playlist` type is set to `unknown`. I first tried setting it to `{ name: string }` but, I was getting poor results with this. Using GPT 3.5, it was attempting to pass the name directly into the play function and not the previously found playlist.

| `{ name: string }` | `undefined` |
| ----- | -----|
| image | image |

### DIFF
--- a/examples/music/src/chatifyActionsSchema.ts
+++ b/examples/music/src/chatifyActionsSchema.ts
@@ -1,6 +1,8 @@
 // This is a schema for writing programs that control a Spotify music player
 
-export type Track = { name: string }
+export type Playlist = unknown;
+
+export type Track = { name: string };
 export type TrackList = Track[];
 
 export type FavoritesTerm =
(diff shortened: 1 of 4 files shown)


steveluc:
Thanks for the contribution, DugganB!  We will experiment with the Playlist type and try changing it from unknown to an array type like TrackList.

PR #50, opened 2023-07-24
Support OpenAI Organization ID

umeshma:
Some Open AI users can belong to multiple organizations within the same account.  This PR adds support for including **OpenAI-Organization** header. 

[https://platform.openai.com/docs/api-reference/requesting-organization](url)

The PR also includes a few additions to .gitignore

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -58,8 +58,8 @@ export function createLanguageModel(env: Record<string, string | undefined>): Ty
  * Creates a language model encapsulation of an OpenAI REST API endpoint.
  * @param apiKey The OpenAI API key.
  * @param model The model name.
- * @param org: The OpenAI organization id.
  * @param endPoint The URL of the OpenAI REST API endpoint. Defaults to "https://api.openai.com/v1/chat/completions".
+ * @param org: The OpenAI organization id.
  * @returns An instance of `TypeChatLanguageModel`.
  */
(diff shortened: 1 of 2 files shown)


steveluc (approved):
This looks good to me.

src/model.ts
```diff
@@ -57,11 +58,17 @@ export function createLanguageModel(env: Record<string, string | undefined>): Ty
  * Creates a language model encapsulation of an OpenAI REST API endpoint.
  * @param apiKey The OpenAI API key.
  * @param model The model name.
+ * @param org: The OpenAI organization id.
  * @param endPoint The URL of the OpenAI REST API endpoint. Defaults to "https://api.openai.com/v1/chat/completions".
  * @returns An instance of `TypeChatLanguageModel`.
  */
-export function createOpenAILanguageModel(apiKey: string, model: string, endPoint = "https://api.openai.com/v1/chat/completions",): TypeChatLanguageModel {
-    return createAxiosLanguageModel(endPoint, { headers: { Authorization: `Bearer ${apiKey}` } }, { model });
+export function createOpenAILanguageModel(apiKey: string, model: string, org: string, endPoint = "https://api.openai.com/v1/chat/completions"): TypeChatLanguageModel {
```
DanielRosenwasser, line 65: Welp, there's the first breaking change!

Maybe we should just switch these into object literals.

umeshma: Good point ... the endpoint was optional, the org was not, so I did I changed the order. Let me make it backward compat.

DanielRosenwasser (approved):
Fine for now- we should just switch to object bags eventually.

PR #59, opened 2023-07-25
Add example of `.env` file

pacexy:
This will make it easier for users to set up environment variables.


lingchuL:
目前代码里是按是否有相应配置来决定要使用OpenAI还是Azure的，并且在检测到有OpenAI的配置会直接使用OpenAI的API尝试，所以如果只是添加用例可能把几行都默认为注释状态比较好~
In the `createLanguageModel` process, which endpoint to use is determined by whether there is a corresponding configuration to them. It will try to get OpenAI env variable first, once it get the `OPENAI_API_KEY`, it will use OpenAI as endpoint. Therefore, it may be better to comment all those lines~

pacexy:
> 目前代码里是按是否有相应配置来决定要使用OpenAI还是Azure的，并且在检测到有OpenAI的配置会直接使用OpenAI的API尝试，所以如果只是添加用例可能把几行都默认为注释状态比较好~ In the `createLanguageModel` process, which endpoint to use is determined by whether there is a corresponding configuration to them. It will try to get OpenAI env variable first, once it get the `OPENAI_API_KEY`, it will use OpenAI as endpoint. Therefore, it may be better to comment all those lines as the default state~

Maybe we can just keep it empty? (e.g. `OPENAI_API_KEY=`), and then it will not affect the `createLanguageModel`.

lingchuL:
Yeah, it works too.

Stevenic:
You don't normally include the "" quotes in `.env.example` files. empty is better

pacexy:
> You don't normally include the "" quotes in `.env.example` files. empty is better

Updated.

robgruen (approved):

PR #65, opened 2023-07-26
Add "TypeChat in 5 Minutes" doc

DanielRosenwasser:
I'm not sure if this is the appropriate title - but I think this sort of doc would be helpful for users who are looking for a breakdown of our sentiment example.

Some TODOs:

- Add the "here's how it looks when put together"
- Make sure it all can be copy-pasted together
- Decide on the right title if not this


site/src/docs/typechat-in-5-minutes.md
```diff
@@ -0,0 +1,147 @@
+---
+layout: doc-page
+title: TypeChat in 5 Minutes
+---
+
+Currently TypeChat's a very small library, so let's start breaking down a few functions to better understand how to use it.
+
+## Providing a Model
+
+TypeChat is mostly model agnostic.
```
pierceboggan, line 10: What does "mostly" mean here?

site/src/docs/typechat-in-5-minutes.md
```diff
@@ -0,0 +1,147 @@
...
+You can populate your environment variables, and based on whether `OPENAI_API_KEY` or `AZURE_OPENAI_API_KEY` is set, you'll get a model of the appropriate type.
+
+```ts
+import dotenv from "dotenv";
+dotenv.config(/*...*/);
+import * as typechat from "typechat";
+const model = typechat.createLanguageModel(process.env);
+```
+
+Regardless, of how you decide to construct your model, we recommend keeping your secret tokens/API keys in a `.env` file, and specifying `.env` in a `.gitignore`.
```
pierceboggan, line 58: Could we provide an example of an alternative model other than OpenAI?

site/src/docs/typechat-in-5-minutes.md
```diff
@@ -0,0 +1,147 @@
...
+
+The first tends to be simpler, so that's what we'll focus on for the remainder of this example.
+
+```ts
+const translator =
+    typechat.createJsonTranslator<SentimentResponse>(model, schema, "SentimentResponse");
+```
+
+A translator takes a model, the schema file contents, and the name of the type within the schema file.
+That type name effectively acts as the "entry-point" of our schema for the language model.
```
pierceboggan, line 95: It might be worth being explicit that your schema can (obviously) contain multiple types, but you must specify the "entry" type.

site/src/docs/typechat-in-5-minutes.md
```diff
@@ -0,0 +1,147 @@
...
+Input strings will be read from this file on a per-line basis.
+If the file name was `undefined`, `processRequests` will work on standard input and provide an interactive prompt.
+Using `process.argv[2]` makes our program interactive by default unless the person running the program provided an input file as a command line argument (e.g. `node ./dist/main.js inputFile.txt`).
+
+Finally, there's the request handler.
+We'll fill that in next.
+
+## Translating Requests
+
+Our handler gets a "request string" each time it's called.
```
pierceboggan, line 126: "request string" = user input?

site/src/docs/typechat-in-5-minutes.md
```diff
@@ -0,0 +1,147 @@
...
+    if (!response.success) {
+        console.log(response.message);
+        return;
+    }
+    console.log(`The sentiment is ${response.data.sentiment}`);
+});
+```
+
+We're calling the `translate` method on each string and getting a response.
+If something went wrong, `response.success` is `false` and we'll be able to grab a `message` explaining what went wrong.
```
pierceboggan, line 141: What happens in the event TypeChat needs to use a repair prompt?

site/src/docs/typechat-in-5-minutes.md
DanielRosenwasser: I was kind of thinking of having a separate page for that with specific examples for each model. We could possibly use a dummy example here?

Underewarrr: > I was kind of thinking of having a separate page for that with specific examples for each model. We could possibly use a dummy example here? What do you think?

I think having an separeted page for this instructions will be amazing!

weykon:
Looking forward to this PR

bennycode:
Thank you @DanielRosenwasser for offering this handy guide on using TypeChat. I was very excited when I first learnt about TypeChat and quickly went through the [examples](https://github.com/microsoft/TypeChat/tree/main/examples) to integrate it with my existing code. 

During the process, I encountered some challenges when using TypeChat within an ESM module. I'd like to share my findings, as they might be helpful in preventing difficulties for others.

I tried to load the schema file but noticed that you cannot simply target it with `__dirname`:

```ts
// Load up the schema file contents.
const schema = fs.readFileSync(path.join(__dirname, "sentimentSchema.ts"), "utf8");
```

After executing the code, I got the following error:

> ReferenceError: __dirname is not defined in ES module scope

I was able to solve this problem by mimicking the behavior of `__dirname` and `__filename` using [import.meta](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Operators/import.meta):

```ts
import fs from 'node:fs';
import path from 'node:path';
import {fileURLToPath} from 'node:url';

const __filename = fileURLToPath(import.meta.url);
const __dirname = path.dirname(__filename);

const texts = path.join(__dirname, 'texts.txt');
const schemaFile = path.join(__dirname, 'sentimentSchema.ts');

// Load up the schema file contents.
const schema = fs.readFileSync(schemaFile, 'utf8');
```

I encountered another issue when trying to import the schema using the following method:

```ts
// Load up the type from our schema.
import type { SentimentResponse } from "./sentimentSchema";
```  

I received a `CustomError` because Node.js couldn't locate the module, even though I was running my code using `ts-node-esm`:

> CustomError: Cannot find module

After doing some research, I found that [file extensions are mandatory](https://nodejs.org/api/esm.html#mandatory-file-extensions) when using ES Modules to resolve relative specifiers. Since `.ts` extensions are only supported with [allowImportingTsExtensions](https://www.typescriptlang.org/tsconfig#allowImportingTsExtensions), I made the decision to use `.js` extensions (even though I was referring to TS code), and it did the trick for me:

```ts
// Load up the type from our schema.
import type { SentimentResponse } from "./sentimentSchema.js";
```

The rest of the description worked perfectly, and I'm very happy about what you and the team are creating here. I'm so happy that I even created a [TypeChat video tutorial](https://www.youtube.com/watch?v=XMoAQSzOlpk), and it's not longer than 5 minutes. So, "TypeChat in 5 Minutes" is a reality! 😊 

Great job! 👍

Sending my best wishes from Berlin!
Benny

DanielRosenwasser:
Thanks @bennycode, this is great feedback - I hadn't yet gotten the chance to work on ESM examples, and I knew that there would be some rough patches with `__dirname`. It'd be ideal for all of our example code to just work across ESM and CJS, but `import.meta` isn't available outside of modules.

hjonasson:
I read through the entire doc and learned a lot! Can/will this PR be completed?

rogerpadilla (approved):

PR #66, opened 2023-07-27
fix: "Restaurant" example has wrong output in README #62

### LINKED ISSUE #62: "Restaurant" example has wrong output in README
The "Restaurant" example has a wrong output in the "Usage" section of the README[^1], if I understand the input correctly, it should be:
```diff
-2 large pizza with mushrooms
+1 large pizza with mushrooms
+1 large pizza with sausage
 1 small pizza with sausage
 1 whole Greek salad
 1 Pale Ale
 1 Mack and Jacks
```

If this is intentional, then maybe there should be a sentence below it explaining this to avoid confusion, for example:
> This shows that TypeChat may not be 100% accurate, and you may want to consider asking the user for confirmation before performing any action. The output here erroneously shows 2 mushroom pizzas and 1 sausage pizza, while it should be 1 mushroom pizza and 2 sausage pizzas (one large and one small).

[^1]: And the "Input" might be incorrect as well, shouldn't it be only `🍕>` instead of `😀> 🍕>`?

weykon:
Fixes #62

### DIFF
--- a/examples/restaurant/README.md
+++ b/examples/restaurant/README.md
@@ -18,18 +18,35 @@ For example, given the following order:
 **Input**:
 
 ```
-🍕> I want three pizzas, one with mushrooms and the other two with sausage.  Make one sausage a small.  And give me a whole Greek and a Pale Ale.  And give me a Mack and Jacks.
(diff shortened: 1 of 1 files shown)


weykon:
the repaired output is re-generated of the origin text.

DanielRosenwasser:
Thank you!

weykon:

PR #84, opened 2023-08-01
Feat: Adding support for dot notation (path) in @ref objects

caridy:
## Feature

@DanielRosenwasser, as discussed in our meeting last week, one of the limitations of TypeChat programs is the ability to use the dot notation to reach into previous results, passing specific bits rather than the entire result object from a previous step. This PR is a very preliminary exploration of this new capability.

## Details

A good way to visualize this is by modifying slightly the `math` example. If the multiplication function returns an object rather than a numeric value, e.g.:

```ts
type MultiplyReturnValue = {
    value: number;
}
```

Then the program produced by the LLM should be slightly different for the following prompt:

```
> multiply two by three, then multiply four by five, then sum the results
```

The output program would look like:

```json
{
  "@steps": [
    {
      "@func": "mul",
      "@args": [
        2,
        3
      ]
    },
    {
      "@func": "mul",
      "@args": [
        4,
        5
      ]
    },
    {
      "@func": "add",
      "@args": [
        {
          "@ref": 0,
          "path": ["value"]
        },
        {
          "@ref": 1,
          "path": ["value"]
        }
      ]
    }
  ]
}
```

You can notice that the `add` operation now uses a different form to obtain the values to be added, it is using the regular `@ref` directive, but also using the `path` definition, which is an optional value to rely on property access mechanism to access bits out of the previous results. (see the PR Diff for more details on the new type).

The `path` is an array to support `string` and `number` in case of arrays. In my experience, producing a string to be parsed makes the LLM a lot more susceptible to errors and hallucinations, at least GPT3.5 some times uses the following `foo.*.bar` which makes no sense. When using an array, it works just fine, e.g.: `["foo", 0, "bar"]`.

As a result, the produced program would look like:

```js
import { API } from "./schema";
function program(api: API) {
  const step1 = api.mul(2, 3);
  const step2 = api.mul(4, 5);
  return api.add(step1.value, step2.value);
}
```

Notice how `api.add` receives specific values from previous result objects.

## Missing Pieces

1. [ ] This only covers the plumbing of arguments, it does not cover the return value. @ahejlsberg during our conversation you talk about the exit mechanism for the program. That is what we use in RS Lang, a standard function that indicates the exit that receives a value (first argument). I don't think we need anything as fancy as that, but perhaps some additional information on the `@func` to indicate the path to be used for the result value, that way the program can use the path in both places, arguments, and steps.
2. [ ] Examples.
3. [ ] Tests.

### DIFF
--- a/src/program.ts
+++ b/src/program.ts
@@ -27,6 +27,8 @@ export type JsonValue = string | number | boolean | null | { [x: string]: Expres
 export type ResultReference = {
     // Index of the previous expression in the "@steps" array
     "@ref": number;
+    // Path to the desired property within the result object, specified as an array of property names or indices
+    "path"?: (string | number)[];
 };
 `;
 
@@ -107,8 +109,24 @@ export function createModuleTextFromProgram(jsonObject: object): Result<string> 
     function objectToString(obj: Record<string, unknown>) {
         if (obj.hasOwnProperty("@ref")) {
             const index = obj["@ref"];
-            if (typeof index === "number" && index < currentStep && Object.keys(obj).length === 1) {
-                return `step${index + 1}`;
+            const hasPath = obj.hasOwnProperty("path");
+            if (typeof index === "number" && index < currentStep && Object.keys(obj).length === (hasPath ? 2 : 1)) {
+                const path = obj["path"];
+                let str = `step${index + 1}`;
+                if (Array.isArray(path)) {
+                    for (const prop of path) {
+                        if (typeof prop === "string") {
+                            str += `.${prop}`;
+                        }
+                        else if (typeof prop === "number") {
+                            str += `[${prop}]`;
+                        }
+                        else {
+                            str += `["${prop}"]`;
+                        }
+                    }
+                }
+                return str;
             }
         }
         else if (obj.hasOwnProperty("@func")) {
@@ -160,7 +178,23 @@ export async function evaluateJsonProgram(program: Program, onCall: (func: strin
         if (obj.hasOwnProperty("@ref")) {
             const index = obj["@ref"];
             if (typeof index === "number" && index < results.length) {
-                return results[index];
+                const path = obj["path"];
+                let value = results[index];
+                if (Array.isArray(path)) {
+                    for (const prop of path) {
+                        if (typeof prop === "string" && typeof value === "object" && value !== null && prop in value) {
+                            value = value[prop as keyof typeof value];
+                        }
+                        else if (typeof prop === "number" && Array.isArray(value) && prop < value.length) {
+                            value = value[prop];
+                        }
+                        else {
+                            return undefined;
+                        }
+                    }
+                    return value;
+                }
+                return value;
             }
         }
         else if (obj.hasOwnProperty("@func")) {


src/program.ts
```diff
@@ -107,8 +109,24 @@ export function createModuleTextFromProgram(jsonObject: object): Result<string>
...
+                if (Array.isArray(path)) {
+                    for (const prop of path) {
+                        if (typeof prop === "string") {
+                            str += `.${prop}`;
+                        }
+                        else if (typeof prop === "number") {
+                            str += `[${prop}]`;
+                        }
+                        else {
+                            str += `["${prop}"]`;
```
caridy, line 125: This part I'm not sure about. The code of the project in general is not using defensive programming, so I kept this simple. IMO, the right thing to do here is to first, have some validation to make sure that prop is safe to be used with the dot notation, otherwise fallback to the `[""]`. At the moment, this else should never be hit by code generated by LLM since it would always be either a string of prop, but it could be that the string is not a valid identifier.

src/program.ts
```diff
@@ -160,7 +178,23 @@ export async function evaluateJsonProgram(program: Program, onCall: (func: strin
...
+                if (Array.isArray(path)) {
+                    for (const prop of path) {
+                        if (typeof prop === "string" && typeof value === "object" && value !== null && prop in value) {
+                            value = value[prop as keyof typeof value];
+                        }
+                        else if (typeof prop === "number" && Array.isArray(value) && prop < value.length) {
+                            value = value[prop];
+                        }
+                        else {
+                            return undefined;
```
caridy, line 192: this needs some consideration. in the prompt template, there is a ref to a function that cannot return `undefined`, which I didn't fully understand the "why". But now with the path, it seems that undefined values are possible.

caridy:

ahejlsberg:
@caridy Thanks for taking the time to put together this experiment.

I've played with it a bit and I'm seeing not seeing a great success rate with the LLMs picking up on the pattern. For example, on the math sample modified to return a `{ value: number }` from each of the API functions, gpt-3.5 pretty consistently fails to add `"path": ["value"]` properties to `@ref` objects. As might be expected, gpt-4 does better, but I worry that's too high of a benchmark.

Also, I'm wondering if we want to endow _every_ program with the ability to access properties on objects without some ability to intervene in the interpreter, in particular because it forces programs to deal with possible `undefined` values.

Finally, and this is more of an implementation detail, with the proposed design it is only possible to access properties through results of previous steps. For example, there's no provision for directly accessing a property of a function result. This forces LLMs to introduce intermediate steps, and we know from experience less capable models like gpt-3.5 have a hard time with that (in fact, this is why we permit nesting of function calls in the first place).

One could argue that property access is just another facet of the design of target APIs. In a sense, there no semantic difference between having properties or having accessor functions in the APIs, but the latter provides a more consistent way of controlling access--it's a function call like any other function call. Indeed, if you want to get fancy, you could even have a generic `getProperty` accessor function in your API.

caridy:
@ahejlsberg let me give it a try to the `getProperty` idea and see how far I can go with that.

As for the no undefined values allowed, that's is the biggest question I have, mostly the reasons behind such decision. Anything that I can read about this topic?

ahejlsberg:
What I was alluding to regarding `undefined` is that the interpreter would need to introduce semantics for properties that are missing (because they're optional) or have `null` or `undefined` values, and it's not clear there's a single answer. You might want exceptions, propagation of the `undefined` value, a default value, etc. So, it quickly gets complicated (just think of the `.` vs `?.` vs `??` operators in JS). It seems a lot more attractive to just make this a facet of the target API by having (or not having) property accessor functions.

caridy:
Ok, tried multiple variants, and I think you are right. It gets complicated very quickly. Thanks for the explanation @ahejlsberg.

Closing this PR. For reference, in case others are facing similar challenges, I used the following function:

```ts
    /** Gets the property of target, equivalent to target[propertyKey]. */
    getProperty: <T extends object, P extends PropertyKey>(inputs: {
        target: T;
        propertyKey: P;
    }) => P extends keyof T ? T[P] : any;
```

It works relatively well, specially on gpt 4, but as discussed above, if you have good types, with well defined functions, it is very likely that you don't need `getProperty`.

PR #88, opened 2023-08-02
Added Promptrix to TypeChat

Stevenic:
Added Promptrix support to TypeChat library and examples.  Translator now returns a `PromptSection` and the model interface has been updated to take `PromptSection` as input along with optional `PromptMemory` and `PromptFunctions` objects which are needed by the renderer. 

This change should put TypeChat in a great position to add support for conversation history and a more chat like experience.

### DIFF
--- a/examples/music/package.json
+++ b/examples/music/package.json
@@ -14,6 +14,7 @@
     "chalk": "^2.3.1",
     "dotenv": "^16.3.1",
     "express": "^4.18.2",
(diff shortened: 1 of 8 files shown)


ahejlsberg:
@Stevenic I don't think we want to add a dependency on another library, but I do agree some more flexibility is needed in this space. I have put up #114 do hopefully address that.

PR #89, opened 2023-08-03
Refactor into a pnpm workspace

### LINKED ISSUE #81: all samples package.json do not have typechat as a dependency
Not sure if this is intentional or not but the `package.json` in each of the samples is missing the dependency for typechat. 

i.e. [https://github.com/microsoft/TypeChat/blob/main/examples/calendar/package.json](https://github.com/microsoft/TypeChat/blob/main/examples/calendar/package.json)

jakebailey:
Fixes #81

Outstanding questions:

- Should it really do `build-all` on `prepare`?
- Should `site` be a part of the monorepo?
	- If so, should `build-all` build the site?
- Should I add a corepack version specifier for pnpm, or will people be able to figure out how to install pnpm themselves?
- Would you benefit from something that helps bump the version numbers in the examples? I'm not sure what this would be other than something like [`changesets`](https://pnpm.io/using-changesets).

Theoretically I can attempt to do this via plain `npm` though @DanielRosenwasser had problems with it hence asking me to send this PR for real.


.github/workflows/ci.js.yml
```diff
@@ -30,10 +30,17 @@ jobs:
...
     - name: Use Node.js ${{ matrix.node-version }}
       uses: actions/setup-node@v3
       with:
         node-version: ${{ matrix.node-version }}
-        cache: 'npm'
-    - run: npm ci
-    - run: npm run build-all
+        cache: 'pnpm'
+
+    - run: pnpm install
```
jakebailey, line 45: This awkwardly already runs `build-all`.

jakebailey:
> No, I want that to stay out of everyone's way.

Do you want me to convert that to `pnpm` just so it's not mixed package managers?

> I don't think it makes much of a difference, right? If we did this, would it work out of the box for anyone with Node 18+? Or would people have to run a command?

People still have to run `corepack enable`, it's just that you will only be using one version of pnpm, which I'd argue is better (this is how I have it set up on TS).

> > Would you benefit from something that helps bump the version numbers in the examples? I'm not sure what this would be other than something like [`changesets`](https://pnpm.io/using-changesets).
> 
> I can just run that manually I think. It doesn't need to be a script.

Right now there's no script or anything; you can technically leave the old specifiers in but pnpm won't help you update them everywhere as far as I know.

README.md
```diff
@@ -27,7 +27,10 @@ npm install typechat
 You can also build TypeChat from source:
 
 ```
-npm run build
+git clone https://github.com/microsoft/TypeChat
+cd TypeChat
+pnpm install
```
axecopfire, line 32: Not everyone knows `pnpm`, I'd recommend adding `pnpm` install instructions (it could seem like a typo. https://pnpm.io/cli/install

PR #90, opened 2023-08-03
Add typechat to example package deps

### LINKED ISSUE #81: all samples package.json do not have typechat as a dependency

jakebailey:
This is an alternative to #89, and also fixes #81.

This worked for me but not for @DanielRosenwasser; I did see a bunch of time spent idealTree-ing but that happens when you change the tree, it seems.

There are other reasons to take #89 instead, mainly, perf.

### DIFF
--- a/examples/calendar/package.json
+++ b/examples/calendar/package.json
@@ -11,7 +11,8 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "dotenv": "^16.3.1"
+    "dotenv": "^16.3.1",
+    "typechat": "^0.0.10"
   },
   "devDependencies": {
     "@types/node": "^20.3.1",
--- a/examples/coffeeShop/package.json
+++ b/examples/coffeeShop/package.json
@@ -11,7 +11,8 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "dotenv": "^16.3.1"
+    "dotenv": "^16.3.1",
+    "typechat": "^0.0.10"
   },
   "devDependencies": {
     "@types/node": "^20.3.1",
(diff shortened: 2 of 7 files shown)


DanielRosenwasser:
It works, but my issue was that if you run with a specific `install-strategy` in an `.npmrc`, that doesn't seem to do anything different than usual. It appears that everything gets hoisted.

jakebailey:
The layouts you get between this PR and the pnpm one are the same; you get hoisted to the top no matter what, so I'm not sure what you're observing that's different/bad exactly.

DanielRosenwasser:
My assumption was that pnpm creates a specific `node_modules` directory per package which ends up being "stricter". The two main things which that helps with is:

1. pnpm will not directly expose a package that an example relies on (like `dotenv`) to `typechat` itself
1. examples that don't explicitly depend on `typechat` will not be able to access it

jakebailey:
It may or may not do either because you're using the monorepo root as an actual package with dependencies. It's impossible for an example to _not_ see `axios`, for example; it's there in the parent directory, and you can't fix that without using the more typical layout which moves everything into `packages/*` or something.

But, it is at least _more unlikely_ that you'll accidentally use a package that hasn't been declared thanks to pnpm not linking anything but declared ones.

In this repo's case, I believe you will get an error if typechat is not listed even though it's a parent because pnpm doesn't (yet) automatically create self-links like npm/yarn do.

PR #108, opened 2023-08-21
examples: fix "main" entry point in package.json

DLehenbauer:
The "main" entry point for the samples previously pointed to the non-existent 'index.js'.

Fixing this allows people to run samples with the shorthand `node .`, which consults ./package.json for the entry point.

Of course, `node ./dist/main.js` still works too.

I did not update the readme or examples.md to use `node .`, since I thought this was less obvious, but can if desired.

### DIFF
--- a/examples/calendar/package.json
+++ b/examples/calendar/package.json
@@ -3,7 +3,7 @@
   "version": "0.0.1",
   "private": true,
   "description": "",
-  "main": "index.js",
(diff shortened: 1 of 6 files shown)


DanielRosenwasser (approved):

DanielRosenwasser:
Great idea - thank you!

PR #114, opened 2023-08-31
Add support for prompt roles and prompt preambles

ahejlsberg:
This PR adds support for prompt roles, i.e. prompt sections tagged with `"system"`, `"user"`, or `"assistant"` roles. TypeChat uses the `"user"` role for prompts it generates and the `"assistant"` role for previous LLM responses (which will be part of the prompt in repair attempts). TypeChat currently doesn't use the `"system"` role.

This PR also adds support for specifying a prompt preamble when processing requests. This makes it easier to implement chat-like interactions and conversation history.

### DIFF
--- a/src/typechat.ts
+++ b/src/typechat.ts
@@ -1,5 +1,5 @@
 import { Result, error } from "./result";
-import { TypeChatLanguageModel } from "./model";
+import { TypeChatLanguageModel, PromptSection } from "./model";
 import { TypeChatJsonValidator, createJsonValidator } from "./validate";
 
 /**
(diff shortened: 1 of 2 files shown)


src/typechat.ts
```diff
@@ -111,7 +114,8 @@ export function createJsonTranslator<T extends object>(model: TypeChatLanguageMo
             if (!attemptRepair) {
                 return error(`JSON validation failed: ${validation.message}\n${jsonText}`);
             }
-            prompt += `${responseText}\n${typeChat.createRepairPrompt(validation.message)}`;
```
umeshma, line 114: We'll be sending the responseText twice. Once as {assistant, responseText} and also as part of the prompt.

DanielRosenwasser (approved):

jonparker (approved):

Githamza:
How can I try it ?  I have installed the last version (v0.0.10) of typechat but in node_modules/typechat, these additions does not exists

Githamza:
Thanks , I have a question , why is the `request` parameter mandatory ? If I would send a promptPreamble with `system` and `user` roles , what should I put in `request` param ? for me `request` content is the `user` role content

olawalejuwonm:
How can this be used please?

kad3nce:
@olawalejuwonm, they key change is here: https://github.com/microsoft/TypeChat/pull/114/files#diff-1ece2a0b089b56f4aa22ef64be65b4305b5e890b560a39d9d12140f9a42d40f4R104

Add a `user` message preamble:
```
translator.translate(data, 'Hello there');
```

Add a system prompt:
```
translator.translate(data, [
  { role: 'system', content: 'You are an expert data anonymizer...' }
]);
```

olawalejuwonm:
How do I pass the user content then?

PR #115, opened 2023-08-31
Add `validateInstance` method to `TypeChatJsonTranslator`

### LINKED ISSUE #91: How to impose additional validation logic?
This library is great, but how can I add additional validation? 
For example, how can I ensure that a string doesn't contain newlines?

GPT3 will often ignore instructions to return a string in a certain format, and it would be great if this library could retry until the right response (e.g. passing validation) was returned.

ahejlsberg:
This PR adds a `validateInstance` method to the `TypeChatJsonTranslator` type. This function is called following successful schema validation of an instance. By default the function just returns a `Success<T>`, but an application can assign a new function that implements any additional validation.

Fixes #91.

### DIFF
--- a/src/typechat.ts
+++ b/src/typechat.ts
@@ -1,4 +1,4 @@
-import { Result, error } from "./result";
+import { Result, success, error } from "./result";
 import { TypeChatLanguageModel } from "./model";
 import { TypeChatJsonValidator, createJsonValidator } from "./validate";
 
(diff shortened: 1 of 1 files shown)


src/typechat.ts
```diff
@@ -42,6 +42,15 @@ export interface TypeChatJsonTranslator<T extends object> {
...
     createRepairPrompt(validationError: string): string;
+    /**
+     * Optionally implements additional validation logic beyond what is expressed in the schema. This function is
+     * called following successful schema validation of an instance. By default the function just returns a
+     * `Success<T>` for the given instance, but an application can assign a new function that implements any
+     * additional validation.
+     * @param instance The instance to validate.
+     * @returns A `Success<T>` with the final validated instance, or an `Error` explaning the validation failure.
+     */
+    validateInstance(instance: T): Result<T>;
```
DanielRosenwasser, line 53: Aside - because you mention that this is "optional", it's a little weird that it's always present.

Also, it is a little odd that this API really expects people to assign directly to the result object. You can't do something like

```ts
const translator = {
  ...createJsonTranslator(/*...*/),
  validateInstance: /*...*/,
}
```

DanielRosenwasser (approved):

src/typechat.ts
ahejlsberg: Well, it's optional to implement it, but that doesn't mean there isn't a default implementation. Right, you can't just spread like that, the methods are bound to the particular instance (i.e. they don't use `this`).

caridy:
This is great. Thanks for adding the additional hook.

PR #135, opened 2023-11-10
Add blog post discussing TypeChat Programs.

DanielRosenwasser:


site/src/blog/typechat-programs.md
```diff
@@ -0,0 +1,302 @@
...
+            fs.writeFileSync(filePath, contents, { encoding: "utf8" });
+        }
+        case "trimString": {
+            const [input] = args as [string];
+            return input.trim();
+        }
+    }
+});
+```
+
```
ahejlsberg, line 217: Here I would talk about the advantages of the callback model: The evaluator *never* calls any functions, that's all up to the callback. And the callback gets to do any amount of validation it deems necessary. In other words, the host is fully in control of what transpires.

site/src/blog/typechat-programs.md
```diff
@@ -0,0 +1,302 @@
...
+    trimString(input) {
+        return input.trim();
+    },
+} satisfies AsyncAPI;
+```
+
+## Early Alternatives
+
+When designing TypeChat Programs, we played around with several different concepts.
+One idea we had was to generate a TypeScript program that could be executed directly.
```
ahejlsberg, line 272: ~TypeScript~ JavaScript

ahejlsberg (commented):
This is great. Thanks for writing this post!

site/src/blog/typechat-programs.md
DanielRosenwasser:
```suggestion
One idea we had was to generate a JavaScript program that could be executed directly.
```

PR #139, opened 2023-11-15
Fix authentication issue when using managed identity on Azure

sinedied:
When using [managed identity](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/managed-identity) on Azure, the current code is not working and returning 401 errors.

The header `Authorization: `Bearer ${apiKey}` needs to be set for that use case.

I can also add a note in the documentation on how to use Azure managed identity if you want, let me know.


ZackaryJacobthereal (commented):
Z.1

PR #146, opened 2023-12-12
Added launch and tasks JSON to enable debugging from root folder.

pcdeadeasy:
This PR enables debugging the examples from the root folder by just hitting F5 and choosing the right configuration. To enable stepping into the typechat code base, you will need to run `npm link` from ~/src.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -954,9 +954,9 @@
       }
     },
     "node_modules/follow-redirects": {
-      "version": "1.15.2",
-      "resolved": "https://registry.npmjs.org/follow-redirects/-/follow-redirects-1.15.2.tgz",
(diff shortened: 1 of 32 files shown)


hillary-mutisya (approved):

DanielRosenwasser:
@davrous had a similar PR over at #40 - but given that we don't know if a user will want to use Node, Python, etc., does this make sense? Do multi-root workspaces help here?

pcdeadeasy:
> @davrous had a similar PR over at #40 - but given that we don't know if a user will want to use Node, Python, etc., does this make sense? Do multi-root workspaces help here?

@DanielRosenwasser, I updated this PR instead of the classification sample. The updated launch.json with console set to integratedTerminal works for both windows and WSL.

PR #147, opened 2023-12-13
Zod validation support, switch to Fetch API, and refactoring

ahejlsberg:
This PR introduces the following changes:

* Support for validation using [Zod](https://zod.dev/), a popular TypeScript-first schema validation library.
* Switch to using the JavaScript Fetch API for LLM access and remove dependency on `axios` package.
* Refactoring of certain functions and module names.

### Zod support

The new Zod support is useful in applications that already incorporate JSON validation and/or dynamically change schema based on interaction history, databases, or other information sources. Also, Zod is significantly lighter weight than incorporating the TypeScript compiler for validation purposes.

Even though validation of JSON responses can now be performed by the Zod validator, schemas are still communicated to the LLM as TypeScript type definitions. The new `typechat/zod` validator automatically converts Zod schema to TypeScript source code suitable for inclusion in LLM prompts (and the conversion is available in a helper function for other purposes). In our experience, using TypeScript to describe schema to LLMs produces better outcomes and consumes fewer tokens than using JSON Schema.

Two new examples, `examples/coffeeShop-zod` and `examples/sentiment-zod`, demonstrate the new Zod support. These are Zod equivalents of the corresponding examples that use the TypeScript compiler for validation.

An example of a simple Zod schema for determining the sentiment of a some user input.:

```ts
import { createLanguageModel, createJsonTranslator, processRequests } from "typechat";
import { createZodJsonValidator } from "typechat/zod";
import { z } from "zod";

const SentimentResponse = z.object({
    sentiment: z.enum(["negative", "neutral", "positive"]).describe("The sentiment of the text")
});

const SentimentSchema = {
    SentimentResponse
};

const model = createLanguageModel(process.env);
const validator = createZodJsonValidator(SentimentSchema, "SentimentResponse");
const translator = createJsonTranslator(model, validator);

// Process requests interactively or from the input file specified on the command line
processRequests("😀> ", undefined, async (request) => {
    const response = await translator.translate(request);
    if (!response.success) {
        console.log(response.message);
        return;
    }
    console.log(`The sentiment is ${response.data.sentiment}`);
});
```

In addition to the `createZodJsonValidator` function, the new `typechat/zod` module exports a convenient `getZodSchemaAsTypeScript` function that can be used to convert Zod schemas into TypeScript source code.

### Switch to Fetch API

The implementation has switched to using the JavaScript Fetch API instead of depending on the `axios` package. This helps reduce bundle sizes with no effect on client applications.

### Refactoring

The implementation has been refactored as follows:

* Validators are now in the `typechat/ts` and `typechat/zod` modules (instead of the main `typechat` module).
* The `createJsonTranslator` function now takes a `validator` parameter instead of actually creating the validator.
* The `TypeChatJsonValidator` interface has been changed to have methods instead of properties for accessing the schema text and target type name, and the `validate` method has been changed to take an `object` instead of a `string` parameter.

These are breaking changes, but quite easy to adopt.

### DIFF
--- a/examples/calendar/src/main.ts
+++ b/examples/calendar/src/main.ts
@@ -2,6 +2,7 @@ import fs from "fs";
 import path from "path";
 import dotenv from "dotenv";
 import { createLanguageModel, createJsonTranslator, processRequests } from "typechat";
+import { createTypeScriptJsonValidator } from "typechat/ts";
 import { CalendarActions } from './calendarActionsSchema';
 
 // TODO: use local .env file.
@@ -9,8 +10,9 @@ dotenv.config({ path: path.join(__dirname, "../../../.env") });
 
 const model = createLanguageModel(process.env);
 const schema = fs.readFileSync(path.join(__dirname, "calendarActionsSchema.ts"), "utf8");
-const translator = createJsonTranslator<CalendarActions>(model, schema, "CalendarActions");
-translator.validator.stripNulls = true;
+const validator = createTypeScriptJsonValidator<CalendarActions>(schema, "CalendarActions");
+const translator = createJsonTranslator(model, validator);
+//translator.stripNulls = true;
 
 // Process requests interactively or from the input file specified on the command line
 processRequests("📅> ", process.argv[2], async (request) => {
--- a/examples/music/package.json
+++ b/examples/music/package.json
@@ -11,18 +11,19 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
+    "axios": "^1.6.2",
     "chalk": "^2.3.1",
     "dotenv": "^16.3.1",
     "express": "^4.18.2",
     "open": "^7.0.4",
     "sqlite3": "^5.1.6",
-    "typechat": "^0.0.10"
+    "typechat": "^1.0.0"
   },
   "devDependencies": {
     "@types/express": "^4.17.17",
-    "@types/node": "^20.3.1",
+    "@types/node": "^20.10.4",
     "@types/spotify-api": "^0.0.22",
     "copyfiles": "^2.4.1",
-    "typescript": "^5.1.3"
+    "typescript": "^5.3.3"
   }
 }
--- a/examples/music/package.json
+++ b/examples/music/package.json
@@ -17,7 +17,7 @@
     "express": "^4.18.2",
     "open": "^7.0.4",
     "sqlite3": "^5.1.6",
-    "typechat": "^1.0.0"
+    "typechat": "^0.1.0"
   },
   "devDependencies": {
     "@types/express": "^4.17.17",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "typechat",
   "author": "Microsoft",
-  "version": "1.0.0",
+  "version": "0.1.0",
   "license": "MIT",
   "description": "TypeChat is an experimental library that makes it easy to build natural language interfaces using types.",
   "keywords": [
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "typechat",
   "author": "Microsoft",
-  "version": "0.0.10",
+  "version": "1.0.0",
   "license": "MIT",
   "description": "TypeChat is an experimental library that makes it easy to build natural language interfaces using types.",
   "keywords": [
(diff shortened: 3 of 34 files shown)


package.json
```diff
@@ -36,11 +48,11 @@
     "SECURITY.md"
   ],
   "dependencies": {
-    "axios": "^1.4.0",
-    "typescript": "^5.1.3"
+    "typescript": "^5.3.3",
+    "zod": "^3.22.4"
```
jakebailey, line 52: Did the optional peer dependency trick not work out? (Happy to help get that working!)

steveluc:
👍

src/typechat.ts
```diff
@@ -64,17 +63,38 @@ export interface TypeChatJsonTranslator<T extends object> {
     translate(request: string, promptPreamble?: string | PromptSection[]): Promise<Result<T>>;
 }
 
+/**
+ * An object that represents a TypeScript schema for JSON objects.
+ */
+export interface TypeChatJsonValidator<T extends object> {
+    /**
+     * Return a string containing TypeScript source code for the validation schema.
```
DanielRosenwasser, line 71: One of the things I did in Python was I completely decoupled the translator and the validator. So here what you could do is create a `ZodTranslator` and a `ZodValidator`. The `ZodTranslator` would know everything it needs for generating TypeScript and crafting the appropriate prompt for a model. The `ZodValidator` would know everything it needs to validating input JSON.

While it does mean that you need a different kind of validator, I think that separation of concerns would make a lot more sense here. It would also decouple TypeChat validators from specifically talking about TypeScript as a schema language.

____

On the flip side, you could take the stance that the Python implementation implicitly couples the default translator to talking about TypeScript as a schema language - which itself might be worth discussing.

DanielRosenwasser (changes requested):

package.json
```diff
@@ -24,8 +24,20 @@
...
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    },
+    "./ts": {
+      "types": "./dist/ts/index.d.ts",
+      "default": "./dist/ts/index.js"
+    },
+    "./zod": {
+      "types": "./dist/zod/index.d.ts",
+      "default": "./dist/zod/index.js"
```
DanielRosenwasser, line 38: This looks right to me, but figure we can check in with @andrewbranch and @jakebailey.

andrewbranch: Looks good, but can be simplified to

```
".": "./dist/index.js",
"./ts": "./dist/ts/index.js",
"./zod: "./dist/zod/index.js"
```

TS knows how to find sibling types.

Also, note that anyone accidentally using `--moduleResolution node10` (still a default for some settings) will not be able to access these subpaths. TS will tell them to import from `"typechat/dist/zod"` even though that won’t work in anything modern at runtime. It’s up to you whether you want to use that as an opportunity to get them to update their settings or provide a mitigation like one of these: https://github.com/andrewbranch/example-subpath-exports-ts-compat

src/typechat.ts
DanielRosenwasser: What was your take on this idea @ahejlsberg?

src/zod/validate.ts
```diff
@@ -0,0 +1,260 @@
...
+            if (parenthesize) append(")");
+        }
+    }
+
+    function appendTypeDefinition(type: z.ZodType) {
+        switch (getTypeKind(type)) {
+            case z.ZodFirstPartyTypeKind.ZodString:
+                return append("string");
+            case z.ZodFirstPartyTypeKind.ZodNumber:
+                return append("number");
```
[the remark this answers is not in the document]
DanielRosenwasser: Currently it doesn't. Maybe that's fine for now.

examples/calendar/src/main.ts
```diff
@@ -2,15 +2,17 @@ import fs from "fs";
...
 // TODO: use local .env file.
 dotenv.config({ path: path.join(__dirname, "../../../.env") });
 
 const model = createLanguageModel(process.env);
 const schema = fs.readFileSync(path.join(__dirname, "calendarActionsSchema.ts"), "utf8");
-const translator = createJsonTranslator<CalendarActions>(model, schema, "CalendarActions");
-translator.validator.stripNulls = true;
+const validator = createTypeScriptJsonValidator<CalendarActions>(schema, "CalendarActions");
+const translator = createJsonTranslator(model, validator);
+//translator.stripNulls = true;
```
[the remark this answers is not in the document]
DanielRosenwasser: Should we drop this?

DanielRosenwasser:
If we're good with the current state of things, can you just pull from `main`, run an `npm install`, and update the `package-lock.json`?

examples/music/package.json
```diff
@@ -11,18 +11,19 @@
...
   "license": "MIT",
   "dependencies": {
+    "axios": "^1.6.2",
     "chalk": "^2.3.1",
     "dotenv": "^16.3.1",
     "express": "^4.18.2",
     "open": "^7.0.4",
     "sqlite3": "^5.1.6",
-    "typechat": "^0.0.10"
+    "typechat": "^1.0.0"
```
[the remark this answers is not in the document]
hillary-mutisya: Yeah, we probably want a 0.1.0 change at most

src/typechat.ts
hillary-mutisya: I prefer to have the validator have all knowledge of schema (including how to construct the schema string). This works much better in a dynamic schema scenario.

hillary-mutisya (approved):

DanielRosenwasser (approved):

PR #154, opened 2024-01-02
Python: Add support for devcontainers and for comments in generated schema

hillary-mutisya:
- Add devcontainer for python and update hatch initialization. This ensures vscode will use the python interpreter in the hatch virtual environment.
- Add support to get comments from annotated fields. Update the coffeeShop examples to use annotation.
- Add helper method to create the typechat model for OpenAI and AzureOpenAI


pcdeadeasy (approved):
Changes look good to me.

PR #155, opened 2024-01-02
Python: Add sentiment, calendar, and restaurant examples

hillary-mutisya:
This change adds examples that use straightforward schema translation: Sentiment, Restaurant and Calendar, Each example has a console demo as well as a jupyter notebook. The demos are ported from the typescript implementation of TypeChat.

### DIFF
--- a/python/examples/calendar/schema.py
+++ b/python/examples/calendar/schema.py
@@ -1,5 +1,8 @@
 from typing import Literal, NotRequired, TypedDict, Annotated
-def Doc(s: str) -> str: return s
+
+
+def Doc(s: str) -> str:
+    return s
 
 
 class UnknownAction(TypedDict):
@@ -11,10 +14,10 @@ class UnknownAction(TypedDict):
     text: Annotated[str, Doc("text typed by the user that the system did not understand")]
 
 
-class EventTimeRange(TypedDict):
-    startTime: NotRequired[str]
-    endTime: NotRequired[str]
-    duration: NotRequired[str]
+class EventTimeRange(TypedDict, total=False):
+    startTime: str
+    endTime: str
+    duration: str
 
 
 class Event(TypedDict):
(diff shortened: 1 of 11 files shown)


python/examples/calendar/schema.py
```diff
@@ -0,0 +1,89 @@
+from typing import Literal, NotRequired, TypedDict, Annotated
+def Doc(s: str) -> str: return s
```
DanielRosenwasser, line 2: This is okay for now - in the future, we can just import https://typing-extensions.readthedocs.io/en/latest/#Doc

python/examples/calendar/schema.py
```diff
@@ -0,0 +1,89 @@
...
+class UnknownAction(TypedDict):
+    """
+    if the user types text that can not easily be understood as a calendar action, this action is used
+    """
+
+    actionType: Literal["Unknown"]
+    text: Annotated[str, Doc("text typed by the user that the system did not understand")]
+
+
+class EventTimeRange(TypedDict):
```
DanielRosenwasser, line 14: I would add `total=True` and drop all the `NotRequired`

See https://docs.python.org/3/library/typing.html#typing.TypedDict

python/examples/calendar/schema.py
```diff
@@ -0,0 +1,89 @@
...
+    startTime: NotRequired[str]
+    endTime: NotRequired[str]
+    duration: NotRequired[str]
```
DanielRosenwasser, line 17:
```suggestion
class EventTimeRange(TypedDict, total=False):
    startTime: str
    endTime: str
    duration: str
```

python/examples/calendar/schema.py
```diff
@@ -0,0 +1,89 @@
...
+    properties used by the requester in referring to an event
+    these properties are only specified if given directly by the requester
+    """
+
+    day: NotRequired[Annotated[str, Doc("date (example: March 22, 2024) or relative date (example: after EventReference)")]]
+    dayRange: NotRequired[Annotated[str, Doc("(examples: this month, this week, in the next two days)")]]
+    timeRange: NotRequired[EventTimeRange]
+    description: NotRequired[str]
+    location: NotRequired[str]
+    participants: NotRequired[list[str]]
```
DanielRosenwasser, line 39:
```suggestion
class EventReference(TypedDict, total=False):
    """
    properties used by the requester in referring to an event
    these properties are only specified if given directly by the requester
    """

    day: Annotated[str, Doc("date (example: March 22, 2024) or relative date (example: after EventReference)")]
    dayRange: Annotated[str, Doc("(examples: this month, this week, in the next two days)")]
    timeRange: EventTimeRange
    description: str
    location: str
    participants: list[str]
```

pcdeadeasy (approved):
Cool, nice to have these examples in python.

python/examples/calendar/schema.py
hillary-mutisya: fixed

PR #156, opened 2024-01-03
Python: Add math example

hillary-mutisya:
This change adds a simple calculator example which uses JsonProgram for its schema. This is similar to the math example in the typescript implementation. Some changes:

1. The JsonProgram validator and translator are included in the example folder. We'll evaluate whether to include them in the main typechat python implementation at a later date.
2. The translator takes an API that exposes the functions a model can call. The example shows two ways of providing this- either as a TypedDict with Callable values (schema.py) or as a Protocol with instance methods (schemaV2.py).


python/examples/math/program.py
```diff
@@ -0,0 +1,166 @@
...
+    },
+)
+
+JsonValue = str | int | float | bool | None | dict[str, "Expression"] | list["Expression"]
+Expression = JsonValue | FunctionCall | ResultReference
+
+JsonProgram = TypedDict("Program", {"@steps": list[FunctionCall]})
+
+
+async def evaluate_json_program(program: Mapping[str, Any], onCall: Callable[[str, list[Any]], Awaitable[Any]]) -> Any:
```
pcdeadeasy, line 72: Can we use Dict[str, Any], I haven't seen Mapping being used as frequently?
Instead of list[Any] can we Change it to **List**[any]. Also for clarity, can we use type hints from the typing module like List, Dict instead of list, dict etc.

python/examples/math/program.py
```diff
@@ -0,0 +1,166 @@
...
+    results: list[Any] = []
+
+    async def evaluate_array(array: list[Any]) -> list[Any]:
+        return await asyncio.gather(*[evaluate_call(e) for e in array])
+
+    async def evaluate_object(expr: FunctionCall):
```
pcdeadeasy, line 78: would be good to define the return types expected for this function like maybe Union[List[str], ... , None]:

python/examples/math/program.py
```diff
@@ -0,0 +1,166 @@
...
+        results.append(await evaluate_call(step))
+
+    if len(results) > 0:
+        return results[-1]
+    else:
+        return None
+
+
+class TypeChatProgramValidator(TypeChatValidator[T]):
+    def __init__(self, py_type: type[T]):
```
pcdeadeasy, line 110: **Type** instead of type

pcdeadeasy (changes requested):
Great to have parity between the py-typechat and typechat 👍

python/examples/math/program.py
hillary-mutisya: As it turns out, the typing module's Dict, List, Set, Tuple, Type and other types are deprecated aliases for their lowercase syblings (https://docs.python.org/3/library/typing.html#deprecated-aliases). 

The Mapping va. Dict difference is also interesting. The Mapping definition is more generic (just needs a type to define __getitem__,__len__,__iter__) from the ABCs (https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes) while the dict type is a more concrete implememtation. The notes I've seen online recommend using a Mapping type unless you strictly need a dict type.

python/examples/math/program.py
DanielRosenwasser: Might be mistaken, but I do think it should be `type`

pcdeadeasy (approved):
Changes look good.

python/examples/math/program.py
```diff
@@ -0,0 +1,165 @@
...
+)
+import collections.abc
+
+T = TypeVar("T", covariant=True)
+
+def Doc(s: str) -> str:
+    return s
+
+
+program_schema_text = '''
```
DanielRosenwasser, line 24: Why not just generate this with `python_type_to_typescript_schema` so you have a single source of truth?

python/examples/math/program.py
```diff
@@ -0,0 +1,165 @@
...
+async def evaluate_json_program(program: JsonProgram, onCall: Callable[[str, Sequence[Expression]], Awaitable[Expression]]) -> Expression | Sequence[Expression]:
+    results: list[Expression] | Expression = []
+
+    async def evaluate_array(array: Sequence[Expression]) -> Sequence[Expression]:
+        return await asyncio.gather(*[evaluate_call(e) for e in array]) # type: ignore
+
+    async def evaluate_object(expr: FunctionCall):
+        if "@ref" in expr:
+            index = expr["@ref"]
+            if index < len(results):
```
DanielRosenwasser, line 81: You probably need to make sure `expr` is an `int`, and report otherwise, and report if it is out of bounds.

python/examples/math/schemaV2.py
DanielRosenwasser, line 1: I'd get rid of schemaV2, it's not used.

python/examples/math/program.py
DanielRosenwasser, line 1: I'd rename this to `program_translator`

DanielRosenwasser (changes requested):

python/examples/math/program.py
hillary-mutisya: Let's sync up on this. As I see it, the goal of the examples is to show devs how to use the library. In this case, the program APIs can be presented using the typeddict-of-callables approach (schema.py) or the protocol approach (schemav2.py).  We support both approaches and the devs are free to choose either one.  I'm going to add some Readme files that will explain how to switch between the different schema definitions

python/examples/math/schemaV2.py
hillary-mutisya: Let's sync up on this. As I see it, the goal of the examples is to show devs how to use the library. In this case, the program APIs can be presented using the typeddict-of-callables approach (schema.py) or the protocol approach (schemav2.py). We support both approaches and the devs are free to choose either one. I'm going to add some Readme files that will explain how to switch between the different schema definitions

PR #157, opened 2024-01-03
Set developer settings when using the devcontainer in python folder

hillary-mutisya:
This change configures hatch to set up the python virtual environment in the project folder when using the dev container. We also set the python interpretter path to the hatch virtual environment.


DanielRosenwasser:
I spoke to @luabud who confirmed that this is going to end up being unreliable across machines. She suggested instead including the settings within the `devcontainer.json` itself:

```json5
    // Configure tool-specific properties.
    "customizations": {
        "vscode": {
            "extensions": [
                "ms-python.python",
            ],
            "settings": {
                "python.defaultInterpreterPath": "/workspaces/TypeChat/python/.hatch/typechat/bin/python3.12"
            }
        }
    }
 
```

What do you think?

hillary-mutisya:
That works! I updated the PR

PR #158, opened 2024-01-03
Python: Add music player example

hillary-mutisya:
This change adds a music player example to the python typechat implementation. Unlike the typescript edition, this example does not rely on JsonProgram. Instead, the music player schema is exposed as a set of actions (and their corresponding parameter types). The model selects a sequence of actions to run. These are passed to a handler function that executes them in order.

The code is organized as follows:
- The *demo.py* file is the console app entry point - this is similar to previous python examples.
- The *schema.py* file defines the action types and their associated parameters.
- The *notebooks/music.pynb* file has an interactive Jupyter notebook for the music scenario.
- The *client.py* file has the logic for executing schmatized actions against the spotify API
- For this example, we are using the popular Spotipy (https://spotipy.readthedocs.io/en/2.22.1/) package to interact with spotfiy's API. This package does not currently provide type hints on its method,, so we have added code in *spotipyWrapper.py* that wraps the methods that we use. This file also has stub implementations of the data contracts we use in the example. Spotipy2 (https://pypi.org/project/spotipy2/) is a promising effort that has type hints and async method definitions - it's currently missing key parts of Spotify automation that we use (e.g. controlling playback on a particular device). ''Once Spotipy2 is ready, we can retire the wrapper code.

### DIFF
--- a/python/examples/music/client.py
+++ b/python/examples/music/client.py
@@ -1,3 +1,9 @@
+import os
+import sys
+current_path = os.path.abspath(os.path.dirname(__file__))
+if current_path not in sys.path:
+    sys.path.append(current_path)
+
 import math
 from typing import Any, Optional
 
--- a/python/examples/music/client.py
+++ b/python/examples/music/client.py
@@ -1,0 +1,480 @@
+import math
+from typing import Any, Optional
+
+from pydantic.dataclasses import dataclass
+from schema import PlayerAction
+import spotipy
+
+
+@dataclass
+class SimplifiedTrackInfo:
+    name: str
+    uri: str
+    artistNames: list[str]
+    artistUris: list[str]
+    albumName: str
+    genres: Optional[list[str]] = None
+
+
+@dataclass
+class SimplifiedPlaylistInfo:
+    name: str
+    id: str
+
+
+# The spotipy library does not provide type hints or async methods.
+# This wrapper class allows the rest of the code to use type hints and async pattern
+class AsyncSpotipy:
+    _service: spotipy.Spotify
+
+    def __init__(self, service: spotipy.Spotify):
+        super().__init__()
+        self._service = service
+
+    async def devices(self) -> dict[str, Any]:
+        return self._service.devices()  # type: ignore
+
+    async def start_playback(
+        self,
+        device_id: str | None = None,
+        context_uri: str | None = None,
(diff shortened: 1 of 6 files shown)


python/examples/music/client.py
```diff
@@ -0,0 +1,378 @@
...
+
+from schema import PlayerAction
+from spotipyWrapper import SimplifiedTrackInfo, SimplifiedPlaylistInfo, AsyncSpotipy
+
+
+class Config:
+    arbitrary_types_allowed = True
+
+
+@dataclass(config=Config)
```
DanielRosenwasser, line 22: Why do you need this to be a Pydantic dataclass? It doesn't look like you serialize it anywhere.

It looks like you just want a regular `dataclass` so you can get an `__init__` for free, right?

python/examples/music/client.py
```diff
@@ -0,0 +1,378 @@
...
+    devices = await context.service.devices()
+    for device in devices["devices"]:
+        if device["is_active"]:
+            print(F"Active Device {device['name']} of type {device['type']}")
+        else:
+            print(f"Device {device['name']} of type {device['type']} is available")
+
+
+async def handle_call(action: PlayerAction, context: ClientContext):
+    match action["actionName"]:
```
DanielRosenwasser, line 187: Something we could do is use a class with a bunch of methods with `hasattr`. I've suggested we could do something similar with a plain object in the TypeScript implementation too.

For now it is probably fine as-is. I wouldn't rewrite everything unless you think you can do it quickly.

DanielRosenwasser (changes requested):
Overall looks promising! I didn't see much else that jumped out at me. I do want to give it a try locally, and I think @steveluc might want to look things over.

steveluc (approved):
Thanks for getting this one ported!  It looks good.  I'd love to try running it locally.  Let's merge and try it out.

python/examples/music/client.py
hillary-mutisya: I've been looking for a way to verify type signatures at runtime. As far as I can tell, the TypedDicts are just dicts at runtime so we cannot "assert x is T" or "cast x as T" to ensure the right properties are present, Let me know if you find a way to do this.

pcdeadeasy (commented):
Looks good to me.

PR #163, opened 2024-01-04
Python: Add multi-schema routing

hillary-mutisya:
This example shows how to use a translator to select a specific schema/agent to run. We register multiple schemas with corresponding descriptions to the first-level classifier. The first-level classifier routes messages to second-level classifiers/translators. 

This example is ported from the TypeChat.Net implementation.


python/examples/multiSchema/router.py
```diff
@@ -0,0 +1,33 @@
+from typing import Any, Callable, Awaitable, Dict
```
DanielRosenwasser, line 1:
```suggestion
from typing import Any, Callable, Awaitable
```

python/examples/multiSchema/router.py
```diff
@@ -0,0 +1,33 @@
+from typing import Any, Callable, Awaitable, Dict
+from typechat import Failure, TypeChatValidator, TypeChatModel
+from translator import TextClassiferTranslator, TextClassification, ClassificationItem
+
+
+class TextRequestRouter:
+    _current_agents: Dict[str, ClassificationItem]
```
DanielRosenwasser, line 7:
```suggestion
    _current_agents: dict[str, ClassificationItem]
```

python/examples/multiSchema/agents.py
```diff
@@ -0,0 +1,106 @@
+import os
+import sys
+
+examples_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", ".."))
+if examples_path not in sys.path:
+    sys.path.append(examples_path)
```
DanielRosenwasser, line 6: Is this just to make sure that you can import from `examples.*`?

DanielRosenwasser (changes requested):

python/examples/multiSchema/agents.py
```diff
@@ -0,0 +1,106 @@
+import os
+import sys
```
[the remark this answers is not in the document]
DanielRosenwasser: I don't have a strong opinion either way because it could be nice to just pass in a function and a string - but if you do choose to go the base class route, the base class should probably make no mention of the privates - `handle_request` and `agent_name` might be fine though.

python/examples/multiSchema/agents.py
hillary-mutisya: yup

hillary-mutisya (commented):
updated to account for PR comments

DanielRosenwasser (approved):

pcdeadeasy (approved):
Changes look good.

PR #164, opened 2024-01-04
Classification example

pcdeadeasy:
Added a new example to support classification using a simple schema file and pre-defined classes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,9 +9,9 @@ node_modules/
 
 # Local development and debugging
 .scratch/
-.vscode
-**/.vscode/*
+#.vscode
+#**/.vscode/*
 **/tsconfig.debug.json
-!**/.vscode/launch.json
+#!**/.vscode/launch.json
 **/build.bat
 
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,240 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "sentiment-wsl",
+            "type": "node",
+            "request": "launch",
+            "program": "${workspaceFolder}/examples/sentiment/src/main.ts",
+            "stopOnEntry": false,
+            "args": [],
+            "cwd": "${workspaceFolder}/examples/sentiment",
+            "skipFiles": [
+                "<node_internals>/**"
+            ],
+            "preLaunchTask": "build-sentiment",
+            "sourceMaps": true,
+            "outFiles": [
+                "${workspaceFolder}/examples/sentiment/dist/**/*.js"
+            ],
+            "console": "integratedTerminal"
+        },
+        {
+            "name": "sentiment",
+            "type": "node",
+            "request": "launch",
+            "program": "${workspaceFolder}/examples/sentiment/src/main.ts",
+            "stopOnEntry": false,
+            "args": [],
+            "cwd": "${workspaceFolder}/examples/sentiment",
+            "skipFiles": [
+                "<node_internals>/**"
+            ],
+            "preLaunchTask": "build-sentiment",
+            "sourceMaps": true,
+            "outFiles": [
+                "${workspaceFolder}/examples/sentiment/dist/**/*.js"
+            ],
+            "console": "externalTerminal"
+        },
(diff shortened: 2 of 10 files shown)


examples/classification/src/classificationSchema.ts
```diff
@@ -0,0 +1,8 @@
+export type Class = {
```
DanielRosenwasser, line 1: Stick to either interfaces or type aliases for object types.

Also, I would not use the name `Class` here.

.gitignore
```diff
@@ -9,9 +9,9 @@ node_modules/
 
 # Local development and debugging
 .scratch/
-.vscode
-**/.vscode/*
+#.vscode
+#**/.vscode/*
```
DanielRosenwasser, line 13: I don't want these removed - if you need an exception for the root `.vscode/launch.json`, I'd add that.

.vscode/launch.json
```diff
@@ -0,0 +1,240 @@
...
+            "sourceMaps": true,
+            "outFiles": [
+                "${workspaceFolder}/examples/coffeeShop/dist/**/*.js"
+            ],
+            "console": "externalTerminal"
+        },
+    ],
+    "compounds": [
+        {
+            "name": "Debug All Projects",
```
DanielRosenwasser, line 223: I don't think this is a realistic task to have. I'd get rid of it.

.vscode/launch.json
```diff
@@ -0,0 +1,240 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "sentiment-wsl",
```
DanielRosenwasser, line 5: Why do we need wsl-specific tasks here? People should just enter the directory from within WSL if that's what they want.

DanielRosenwasser (changes requested):

examples/classification/src/classificationSchema.ts
DanielRosenwasser: Maybe just call it `Classification`

.vscode/launch.json
pcdeadeasy: It's makes it easier to debug from the root. You need a separate config for WSL if you are debugging in that environment using VSCODE.

pcdeadeasy: Yes

examples/classification/src/main.ts
```diff
@@ -0,0 +1,43 @@
...
+dotenv.config({ path: path.join(__dirname, "../../../.env") });
+
+const model = createLanguageModel(process.env);
+const schema = fs.readFileSync(path.join(__dirname, "classificationSchema.ts"), "utf8");
+const translator = createJsonTranslator<ClassificationResponse>(model, schema, "ClassificationResponse");
+
+// Process requests interactively or from the input file specified on the command line
+processRequests("🤗> ", process.argv[2], async (request) => {
+    const initClasses: string = JSON.stringify(classes, undefined);
+    const fullRequest: string = `Classify "${request}" using the following classification table:\n${initClasses}\n`;
```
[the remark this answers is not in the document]
pcdeadeasy: The way I look at it is the sentiment example is a special case of the classifier example. Another way of doing this would be to use a custom translator but we do not need it as the translate method has a `promptPreamble` parameter.

Yes, I could also describe it in the schema as well but that idea is to keep the data (classes) separate from the schema. What do you think?

.vscode/launch.json
DanielRosenwasser: So I tried this out locally - my experience is that `externalTerminal` didn't seem to work at all, regardless of whether I was running in WSL or not. Instead, only `integratedTerminal` worked across both.

So I think that matches what I recommended elsewhere - that all the WSL-oriented tasks should be the defaults, and there shouldn't be any sort of variant. If that's not matching your experience, we can jump on a call and run through it.

pcdeadeasy: Yes, let’s sync on Monday. I want to check the configuration on windows.

robgruen:
Stale PR.  If you want this work in please open a new PR with the latest code changes.

PR #166, opened 2024-01-05
Adding healthData sample for TypeChat

pcdeadeasy:
This is based on the healthData .Net examples. The code works best with gpt-4.

### DIFF
--- a/examples/healthData/README.md
+++ b/examples/healthData/README.md
@@ -1,0 +1,29 @@
+# Health Data Agent
+
+This example requires GPT-4.
+
+Demonstrates a ***strongly typed*** Chat: a natural language interface for entering health information. You work with a Health Data Agent to interactively enter your medications or conditions.
+
+The Health Data Agent shows how strongly typed **Agents with history** could interact with a user to collect information needed for one or more data types ("form filling").
+
+## Target models
+
+For best and consistent results, use **gpt-4**.
+
+## Try the Health Data Agent
+
+To run the Sentiment example, follow the instructions in the [examples README](../README.md#step-1-configure-your-development-environment).
+
+## Usage
+
+Example prompts can be found in [`input.txt`](input.txt).
+
+For example, given the following input statement:
+
+**Input**:
+
+```
+💉💊🤧> klaritin
+```
+
+**Output**:
--- a/examples/healthData/README.md
+++ b/examples/healthData/README.md
@@ -22,8 +22,9 @@ For example, given the following input statement:
 
 **Input**:
 
-```
+```console
 💉💊🤧> klaritin
+
 ```
 
 **Output**:
--- a/examples/healthData/README.md
+++ b/examples/healthData/README.md
@@ -2,9 +2,9 @@
 
 This example requires GPT-4.
 
-Demonstrates a ***strongly typed*** Chat: a natural language interface for entering health information. You work with a Health Data Agent to interactively enter your medications or conditions.
+Demonstrates a ***strongly typed*** chat: a natural language interface for entering health information. You work with a *health data agent* to interactively enter your medications or conditions.
 
-The Health Data Agent shows how strongly typed **Agents with history** could interact with a user to collect information needed for one or more data types ("form filling").
+The Health Data Agent shows how strongly typed **agents with history** could interact with a user to collect information needed for one or more data types ("form filling").
 
 ## Target models
 
--- a/examples/healthData/README.md
+++ b/examples/healthData/README.md
@@ -16,7 +16,7 @@ To run the Sentiment example, follow the instructions in the [examples README](.
 
 ## Usage
 
-Example prompts can be found in [`input.txt`](input.txt).
+Example prompts can be found in [`input.txt`](./src/input.txt).
 
 For example, given the following input statement:
 
--- a/examples/healthData/README.md
+++ b/examples/healthData/README.md
@@ -23,7 +23,7 @@ For example, given the following input statement:
 **Input**:
 
 ```console
-💉💊🤧> klaritin
+🤧> I am taking klaritin for my allergies
 
 ```
 
--- a/examples/healthData/package.json
+++ b/examples/healthData/package.json
@@ -1,0 +1,22 @@
+{
+  "name": "healthdata",
+  "version": "0.0.1",
+  "private": true,
+  "description": "",
+  "main": "dist/main.js",
+  "scripts": {
+    "build": "tsc -p src",
+    "postbuild": "copyfiles -u 1 src/**/*Schema.ts src/**/*.txt dist"
+  },
+  "author": "",
+  "license": "MIT",
+  "dependencies": {
+    "dotenv": "^16.3.1",
+    "typechat": "^0.0.10"
+  },
+  "devDependencies": {
+    "@types/node": "^20.3.1",
+    "copyfiles": "^2.4.1",
+    "typescript": "^5.1.3"
+  }
+}
--- a/examples/healthData/package.json
+++ b/examples/healthData/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "healthdata",
+  "name": "health-data",
   "version": "0.0.1",
   "private": true,
   "description": "",
--- a/examples/healthData/src/healthDataSchema.ts
+++ b/examples/healthData/src/healthDataSchema.ts
@@ -5,7 +5,7 @@ export interface HealthDataResponse {
     data?: HealthData;
     // Use this to ask questions and give pertinent responses
     message?: string;
-    // Use this parts of the user request not translateed, off topic, etc
+    // Use this parts of the user request not translated, off topic, etc.
     notTranslated?: string;
 }
 
--- a/examples/healthData/src/healthDataSchema.ts
+++ b/examples/healthData/src/healthDataSchema.ts
@@ -1,7 +1,8 @@
 // The following is a schema definition for enetring health data.
 
 export interface HealthDataResponse {
-    // Return this if JSON has ALL required information. Else ask questions
+    // ONLY present when ALL required information is known.
+    // Otherwise, use 'message' to keep asking questions.
     data?: HealthData;
     // Use this to ask questions and give pertinent responses
     message?: string;
--- a/examples/healthData/src/healthDataSchema.ts
+++ b/examples/healthData/src/healthDataSchema.ts
@@ -48,7 +48,7 @@ export interface OtherHealthData {
 export interface ApproxQuantity {
     // Default: Unknown. Required
     displayText: string;
-    // Optional: only if precise quantities are available
+    // Only specify if precise quantities are available
     quantity?: Quantity;
 }
 
--- a/examples/healthData/src/healthDataSchema.ts
+++ b/examples/healthData/src/healthDataSchema.ts
@@ -53,7 +53,7 @@ export interface ApproxQuantity {
 }
 
 export interface ApproxDatetime {
-    // Default: Unknown. Required
+    // Default: "unknown"
     displayText: string;
     // If precise timestamp can be set
     timestamp?: string;
--- a/examples/healthData/src/healthDataSchema.ts
+++ b/examples/healthData/src/healthDataSchema.ts
@@ -62,6 +62,6 @@ export interface ApproxDatetime {
 export interface Quantity {
     // Exact number
     value: number;
-    // UNITS include mg, kg, cm, pounds, liter, ml, tablet, pill, cup, per-day, per-week..ETC
+    // Units like mg, kg, cm, pounds, liter, ml, tablet, pill, cup, per-day, per-week, etc.
     units: string;
 }
--- a/examples/healthData/src/healthDataSchema.ts
+++ b/examples/healthData/src/healthDataSchema.ts
@@ -1,4 +1,4 @@
-// The following is a schema definition for ordering lattes.
+// The following is a schema definition for enetring health data.
 
 export interface HealthDataResponse {
     // Return this if JSON has ALL required information. Else ask questions
--- a/examples/healthData/src/healthDataSchema.ts
+++ b/examples/healthData/src/healthDataSchema.ts
@@ -31,7 +31,7 @@ export interface Medication {
 export interface Condition {
     // Fix any spelling mistakes, especially phonetic spelling
     name: string;
-    // When the condition started? Required
+    // When the condition started.
     startDate: ApproxDatetime;
     // Always ask for current status of the condition
     status: "active" | "recurrence" | "relapse" | "inactive" | "remission" | "resolved" | "unknown";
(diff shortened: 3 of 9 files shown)


examples/healthData/src/healthDataSchema.ts
```diff
@@ -0,0 +1,67 @@
+// The following is a schema definition for ordering lattes.
```
hillary-mutisya, line 1: update this for health scenario

hillary-mutisya (approved):

examples/healthData/README.md
```diff
@@ -0,0 +1,30 @@
+# Health Data Agent
+
+This example requires GPT-4.
+
+Demonstrates a ***strongly typed*** Chat: a natural language interface for entering health information. You work with a Health Data Agent to interactively enter your medications or conditions.
+
+The Health Data Agent shows how strongly typed **Agents with history** could interact with a user to collect information needed for one or more data types ("form filling").
```
DanielRosenwasser, line 7:
```suggestion
Demonstrates a ***strongly typed*** chat: a natural language interface for entering health information. You work with a *health data agent* to interactively enter your medications or conditions.

The Health Data Agent shows how strongly typed **agents with history** could interact with a user to collect information needed for one or more data types ("form filling").
```

DanielRosenwasser (changes requested):

examples/healthData/src/translator.ts
```diff
@@ -0,0 +1,75 @@
...
+    source: "system" | "user" | "assistant";
+    body: string;
+};
+
+export interface TranslatorWithHistory<T extends object> {
+    _chat_history: ChatMessage[];
+    _max_prompt_length: number;
+    _additional_agent_instructions: string;
+    translator: TypeChatJsonTranslator<T>;
+    translate(request: string): Promise<Result<T>>;
```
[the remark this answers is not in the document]
pcdeadeasy: Good catch fixed it!

examples/healthData/src/healthDataSchema.ts
pcdeadeasy: Thanks for pointing it out!

examples/healthData/src/translator.ts
```diff
@@ -0,0 +1,75 @@
...
+        const response = await translator.translate(request);
+        if (response.success) {
+            _chat_history.push({ source: 'assistant', body: JSON.stringify(response.data) });
+        }
+        return response;
+    }
+
+    function _create_request_prompt(intent: string): string {
+        // TODO: drop history entries if we exceed the max_prompt_length
+        const history_str = JSON.stringify(_chat_history, undefined, 2);
```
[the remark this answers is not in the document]
pcdeadeasy: Yes I was wondering which module to use in TS for indent/dedent functionality.

examples/healthData/src/healthDataSchema.ts
```diff
@@ -0,0 +1,67 @@
...
+    // Default: Unknown. Required
+    displayText: string;
+    // If precise timestamp can be set
+    timestamp?: string;
+}
+
+export interface Quantity {
+    // Exact number
+    value: number;
+    // UNITS include mg, kg, cm, pounds, liter, ml, tablet, pill, cup, per-day, per-week..ETC
```
DanielRosenwasser, line 65:
```suggestion
    // Units like mg, kg, cm, pounds, liter, ml, tablet, pill, cup, per-day, per-week, etc.
```

examples/healthData/src/healthDataSchema.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export interface ApproxQuantity {
+    // Default: Unknown. Required
+    displayText: string;
+    // Optional: only if precise quantities are available
+    quantity?: Quantity;
+}
+
+export interface ApproxDatetime {
+    // Default: Unknown. Required
```
DanielRosenwasser, line 56:
```suggestion
    // Default: "unknown"
```

examples/healthData/src/healthDataSchema.ts
```diff
@@ -0,0 +1,67 @@
...
+// Use for health data that match nothing else. E.g. immunization, blood prssure etc
+export interface OtherHealthData {
+    text: string;
+    when?: ApproxDatetime;
+}
+
+export interface ApproxQuantity {
+    // Default: Unknown. Required
+    displayText: string;
+    // Optional: only if precise quantities are available
```
DanielRosenwasser, line 51:
```suggestion
    // Only specify if precise quantities are available
```

examples/healthData/src/healthDataSchema.ts
DanielRosenwasser: What does it mean if there's a default but it's required? Why does the LLM have to fill it in?

examples/healthData/src/healthDataSchema.ts
```diff
@@ -0,0 +1,67 @@
...
+    frequency: ApproxQuantity;
+    // E.g. 50 mg. Required
+    strength: ApproxQuantity;
+}
+
+// Disease, Ailment, Injury, Sickness
+export interface Condition {
+    // Fix any spelling mistakes, especially phonetic spelling
+    name: string;
+    // When the condition started? Required
```
DanielRosenwasser, line 34: Do we really need to specify "required" or "optional"? It should fall out of the declaration syntax, right?

```suggestion
    // When the condition started.
```

examples/healthData/src/healthDataSchema.ts
```diff
@@ -0,0 +1,67 @@
+// The following is a schema definition for enetring health data.
+
+export interface HealthDataResponse {
+    // Return this if JSON has ALL required information. Else ask questions
+    data?: HealthData;
+    // Use this to ask questions and give pertinent responses
+    message?: string;
+    // Use this parts of the user request not translateed, off topic, etc
```
DanielRosenwasser, line 8:
```suggestion
    // Use this parts of the user request not translated, off topic, etc.
```

examples/healthData/src/healthDataSchema.ts
```diff
@@ -0,0 +1,67 @@
+// The following is a schema definition for enetring health data.
+
+export interface HealthDataResponse {
+    // Return this if JSON has ALL required information. Else ask questions
```
DanielRosenwasser, line 4:
```suggestion
    // ONLY present when ALL required information is known.
    // Otherwise, use 'message' to keep asking questions.
```

examples/healthData/src/main.ts
```diff
@@ -0,0 +1,51 @@
...
+import path from "path";
+import dotenv from "dotenv";
+import { createHealthDataTranslator } from "./translator";
+import { createLanguageModel, processRequests } from "typechat";
+import { HealthDataResponse } from "./healthDataSchema";
+
+// TODO: use local .env file.
+dotenv.config({ path: path.join(__dirname, "../../../.env") });
+
+const healthInstructions:string = `
```
DanielRosenwasser, line 11: You shouldn't need to annotate this

```suggestion
const healthInstructions = `
```

examples/healthData/src/main.ts
```diff
@@ -0,0 +1,51 @@
...
+        console.log(`Context: ${response.message}`);
+    }
+    else {
+        const healthdata = response.data;
+        console.log("Translation Succeeded! ✅\n");
+        console.log("JSON View");
+        console.log(JSON.stringify(healthdata, undefined, 2));
+
+        const message = healthdata["message"];
+        const notTranslated = healthdata["notTranslated"];
```
DanielRosenwasser, line 41:
```suggestion
        const healthData = response.data;
        console.log("Translation Succeeded! ✅\n");
        console.log("JSON View");
        console.log(JSON.stringify(healthData, undefined, 2));

        const message = healthData.message;
        const notTranslated = healthData.notTranslated;
```

examples/healthData/src/translator.ts
```diff
@@ -0,0 +1,75 @@
...
+    ${historyStr}
+
+    user:
+    The following is a user request:
+    '''
+    ${intent}
+    '''
+    The following is the user request translated into a JSON object with 2 spaces of indentation and no properties with the value undefined:
+    """
+    `;
```
DanielRosenwasser, line 72:
```suggestion
user: You are a service that translates user requests into JSON objects of type "${typename}" according to the following TypeScript definitions:
'''
${schema}
'''

user:
Use precise date and times RELATIVE TO CURRENT DATE: ${now.toLocaleDateString()} CURRENT TIME: ${now.toTimeString().split(' ')[0]}
Also turn ranges like next week and next month into precise dates

user:
${_additionalAgentInstructions}

system:
IMPORTANT CONTEXT for the user request:
${historyStr}

user:
The following is a user request:
'''
${intent}
'''
The following is the user request translated into a JSON object with 2 spaces of indentation and no properties with the value undefined:
"""
`;
```

examples/healthData/src/translator.ts
```diff
@@ -0,0 +1,75 @@
...
+        _translator,
+        translate,
+    };
+
+    return customtranslator;
+
+    async function translate(request: string): Promise<Result<T>> {
+        const response = await _translator.translate(request);
+        if (response.success) {
+            _chatHistory.push({ source: 'assistant', body: JSON.stringify(response.data) });
```
DanielRosenwasser, line 37: You probably didn't mean to stringify the body, since you'll end up stringifying the string version when you generate `historyStr`.

```suggestion
            _chatHistory.push({ source: "assistant", body: response.data });
```

examples/healthData/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "healthdata",
```
DanielRosenwasser, line 2: I can't quite imagine where this is going to be a problem, but the fact that the containing folder has an a name that is distinct by casing on Unix, but not on Windows, might be a problem long term. Let's just call it `health-data` in the meantime?

```suggestion
  "name": "health-data",
```

then re-run `npm install` from the root and update the `package-lock.json`

DanielRosenwasser (changes requested):
This is almost good to go! I had a few suggestions and questions that I do want to get resolved.

examples/healthData/src/healthDataSchema.ts
pcdeadeasy: The field is filled as unknown by the LLM.

examples/healthData/src/translator.ts
```diff
@@ -34,7 +34,7 @@ export function createHealthDataTranslator<T extends object>(model: TypeChatLang
     async function translate(request: string): Promise<Result<T>> {
         const response = await _translator.translate(request);
         if (response.success) {
-            _chatHistory.push({ source: 'assistant', body: JSON.stringify(response.data) });
+            _chatHistory.push({ source: "assistant", body: response.data });
```
pcdeadeasy: @DanielRosenwasser, this needs to be the JSON stringified.

pcdeadeasy:
> This is almost good to go! I had a few suggestions and questions that I do want to get resolved.

Cool, I have resolved all of the suggestions.

examples/healthData/src/healthDataSchema.ts
DanielRosenwasser: Kind of weird - seems like it would be better off as optional entirely in that case

examples/healthData/src/translator.ts
```diff
@@ -0,0 +1,75 @@
...
+    translate(request: string): Promise<Result<T>>;
+}
+
+export function createHealthDataTranslator<T extends object>(model: TypeChatLanguageModel, schema: string, typename: string, additionalAgentInstructions: string): TranslatorWithHistory<T> {
+    const _chatHistory: ChatMessage[] = [];
+    const _maxPromptLength = 2048;
+    const _additionalAgentInstructions = additionalAgentInstructions;
+    
+    const _translator = createJsonTranslator<T>(model, schema, typename);
+    _translator.createRequestPrompt = _create_request_prompt;
```
DanielRosenwasser, line 22:
```suggestion
    _translator.createRequestPrompt = createRequestPrompt;
```

examples/healthData/src/translator.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    async function translate(request: string): Promise<Result<T>> {
+        const response = await _translator.translate(request);
+        if (response.success) {
+            _chatHistory.push({ source: "assistant", body: JSON.stringify(response.data) });
+        }
+        return response;
+    }
+
+    function _create_request_prompt(intent: string): string {
```
DanielRosenwasser, line 42:
```suggestion
    function createRequestPrompt(intent: string): string {
```

examples/healthData/src/translator.ts
```diff
@@ -0,0 +1,75 @@
...
+    const _additionalAgentInstructions = additionalAgentInstructions;
+    
+    const _translator = createJsonTranslator<T>(model, schema, typename);
+    _translator.createRequestPrompt = _create_request_prompt;
+    
+    const customtranslator: TranslatorWithHistory<T> = {
+        _chatHistory,
+        _maxPromptLength,
+        _additionalAgentInstructions,
+        _translator,
```
DanielRosenwasser, line 28: It's still not very canonical to have these sorts of properties, but it's example code so it's probably fine.

examples/healthData/src/translator.ts
```diff
@@ -0,0 +1,75 @@
...
+        translate,
+    };
+
+    return customtranslator;
+
+    async function translate(request: string): Promise<Result<T>> {
+        const response = await _translator.translate(request);
+        if (response.success) {
+            _chatHistory.push({ source: "assistant", body: JSON.stringify(response.data) });
```
DanielRosenwasser, line 37: Can you explain to me why the body needs to be `stringified`?

For example, let's say your response was something like

```json
{ "foo": "bar" }
```

Your initial chat history would be represented in code as:

```json
[
    "{ \"foo\": \"bar\" }"
]
```

and eventually *that* gets stringified when you construct `historyStr`. So what's the rationale?

examples/healthData/src/healthDataSchema.ts
pcdeadeasy: I agree but in this we want the field. Anyways its consistent across both TS and Python.

examples/healthData/src/translator.ts
pcdeadeasy: yeah it's just an example!

examples/healthData/src/translator.ts
pcdeadeasy: type ChatMessage = {
    source: "system" | "user" | "assistant";
    body: object;
};

this fixes it.

examples/healthData/README.md
```diff
@@ -0,0 +1,30 @@
...
+For example, given the following input statement:
+
+**Input**:
+
+```console
+🤧> I am taking klaritin for my allergies
+
+```
+
+**Output**:
```
DanielRosenwasser, line 30: Probably should have something here for consistency with others.

DanielRosenwasser (approved):

PR #167, opened 2024-01-06
Switch to `%pip install` in notebooks instead of `!pip install`.

DanielRosenwasser:
When trying the notebook examples with the Jupyter extension, I get the following error:

> Use '%pip install' instead of '!pip install'
> &nbsp;&nbsp;'!pip install' could install packages to the wrong environment. [More info](https://github.com/microsoft/vscode-jupyter/wiki/Installing-Python-packages-in-Jupyter-Notebooks)

It's documented over here: https://github.com/microsoft/vscode-jupyter/wiki/Installing-Python-packages-in-Jupyter-Notebooks

I'm not sure if we should be using `pip` within notebooks if we already use a package manager, but it I'm inclined to do what the error message says. 😄

(P.S. thanks to @DonJayamanne and @rchiodo for https://github.com/microsoft/vscode-jupyter/issues/6864)


pcdeadeasy (approved):
Changes look good to me.

hillary-mutisya (approved):
good catch

DonJayamanne:
> I'm not sure if we should be using pip within notebooks if we already use a package manager, but it I'm inclined to do what the error message says

Unfortunately there's no generic command that runs for different package managers.
Thus you have to decide and use either %pip or %conda,

PR #168, opened 2024-01-08
Add support for translating dataclasses

### LINKED ISSUE #160: Add `@dataclass` support to Python
Currently, our Python implementation only supports `TypedDicts`. This is okay, but I think we need to support dataclasses.

Given a reference to `Event` for the following code,

```py
from dataclasses import dataclass, field

@dataclass
class EventTimeRange:
    ...

@dataclass
class Event:
    day: str
    time_range: EventTimeRange
    description: str
    location: str | None
    participants: list[str] = field(default_factory=list)
```

We should generate the following schema for `Event`:

```ts
interface Event {
    day: string;
    time_range: EventTimeRange;
    description: string;
    location: string | null;
    participants: string[];
}
```

I've adapted this example from the calendar example. Note two things:

1. The presence of a default makes a field optional, but does not necessarily make it `null`-able.
2. The presence of `| None` in an annotation makes a field `null`-able, but not necessarily optional.

DanielRosenwasser:
This change enables translation of dataclasses into TypeScript schemas. Using dataclasses allows developers to operate over actual Python objects - in contrast, if we took an approach where we only permitted `TypedDict`s, developers would be restricted to only operating over dictionaries. With this, schemas can be generated from one, the other, or a mix of both.

Fixes #160

### DIFF
--- a/python/src/typechat/_internal/ts_conversion/python_type_to_ts_nodes.py
+++ b/python/src/typechat/_internal/ts_conversion/python_type_to_ts_nodes.py
@@ -53,8 +53,14 @@ from typechat._internal.ts_conversion.ts_type_nodes import (
 
 
 class GenericAliasish(Protocol):
-    __origin__: type
-    __args__: tuple[type, ...]
+    __origin__: object
+    __args__: tuple[object, ...]
+
+
+class Annotatedish(Protocol):
+    # NOTE: `__origin__` here refers to `SomeType` in `Annnotated[SomeType, ...]`
+    __origin__: object
+    __metadata__: tuple[object, ...]
 
 
 # type[TypedDict]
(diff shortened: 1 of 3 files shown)


python/src/typechat/_internal/ts_conversion/python_type_to_ts_nodes.py
```diff
@@ -234,70 +246,93 @@ def convert_to_type_node(py_type: object) -> TypeNode:
...
+        If 'optionality_default' is 
+        """
+        current_annotation: object = py_annotation
+        origin: object
+        optional: bool | None = None
+        comment: str | None = None
+        while origin := get_origin(current_annotation):
+            if origin is Annotated and comment is None:
+                current_annotation = cast(Annotatedish, current_annotation)
+                annotation_metadata = current_annotation.__metadata__[0]
```
hillary-mutisya, line 261: There could be multiple annotations on a type and the comment not guaranteed to be at index 0.  The previous implementation looked through all the metadata entries and found the first one that was either a Doc or a str. We probably want to keep that behavior

DanielRosenwasser: It's super questionable to have more than one piece of metadata, but sure, I can go back to finding the first.

PR #169, opened 2024-01-08
Support abstract collection types in Python

DanielRosenwasser:
This change is part clean-up, part feature addition.

Now that https://github.com/microsoft/pyright/issues/6582 is fixed, we can use all the collection types as values and support them as annotations.

Now that https://github.com/microsoft/pyright/issues/6587 is fixed, we can directly reference `Self`.


hillary-mutisya (approved):
looks good

hillary-mutisya (approved):

PR #174, opened 2024-01-23
Make typescript and zod optional peer dependencies

jakebailey:
#147 added zod support, and then moved both zod and typescript support into separate modules / declared exports. The only way to access these are via importing `typechat/ts` and `typechat/zod`.

This PR makes those packages optional peer dependencies. This means that those depending on typechat don't need to install the dependencies for any of the entrypoints they aren't using.

For example, if you're using the zod provider, you will necessarily have installed zod in your package as well in order to construct a schema. But, if you haven't and are instead using typescript, then zod won't be installed.

If someone forgets a dep, they'll get a compiler error complaining about a missing import, so that case should be pretty clear.


DanielRosenwasser:
> If someone forgets a dep, they'll get a compiler error complaining about a missing import, so that case should be pretty clear.

It does mean we'll have to be careful not to re-export either from the top-level entry point, as that would introduce errors, right?

jakebailey:
Yes, correct.

jakebailey:
Though, given most people run skipLibCheck, they may not. But, they will still have needed to either install zod anyway to even construct the schema, and the chances that they _won't_ have typescript installed for the other case seem super slim.

DanielRosenwasser:
We like this in theory, but we do need all the examples to have explicit dependencies on `typescript` rather than listing it in `devDependencies`.

Additionally, we need to decide on which examples use Zod vs TypeScript.

jakebailey:
Is the intent that the examples can be fully extracted out? As it stands now, they get typescript from the repo root.

DanielRosenwasser:
> Is the intent that the examples can be fully extracted out? 

Yeah, it's that and also general guidance on how the examples should be built.

jakebailey:
Currently they devDep; I'll upgrade the ones using TS when I get a chance.

DanielRosenwasser (approved):

PR #175, opened 2024-01-23
Remove indent/dedent, fix up some type-checking errors.

DanielRosenwasser:
The `indent`/`dedent` logic was fragile to refactoring, so I've just moved the string contents over to the gutter.

Additionally, there were a few type-checking issues that were introduced in the samples that I've cleaned up.


hillary-mutisya (approved):
Looks good

pcdeadeasy (approved):
Looks good.

PR #176, opened 2024-01-29
AI Provider agnostic using Langchain

demosjarco:
I started migrating the hard coded openai and azure openai to instead take a langchain model so that this can be provider agnostic.

While the langchain implementation itself works (I tested it by copying the `src` folder into another project and using it), this project has an immense amount of ESModule configuration issues that prevents it from building. I'm not familiar to the extent of this projects ES implementation and honestly I need help on that part.

Lastly, in OpenAI/Azure OpenAI, this uses the json response mode to guarantee JSON response format, but still uses the existing validator.


demosjarco:
@microsoft-github-policy-service agree company="ChainFuse"

PR #178, opened 2024-02-08
Update test cases to work on 3.11

hillary-mutisya:
- use the pre-3.11 approach for declaring generic classes and typealiastype
- check for both typing.TypedDict and typing_extensions.TypedDict in _KNOWN_SPECIAL_BASES

### DIFF
--- a/python/tests/generic_alias_1.py
+++ b/python/tests/generic_alias_1.py
@@ -1,18 +1,22 @@
-from typing import Literal, TypedDict
+from typing import Literal, TypedDict, TypeVar, Generic
+from typing_extensions import TypeAliasType
 
 from typechat import python_type_to_typescript_schema
 
-class First[T](TypedDict):
+T = TypeVar("T", covariant=True)
+
+
+class First(Generic[T], TypedDict):
     kind: Literal["first"]
     first_attr: T
 
 
-class Second[T](TypedDict):
+class Second(Generic[T], TypedDict):
     kind: Literal["second"]
     second_attr: T
 
 
-type FirstOrSecond[T] = First[T] | Second[T]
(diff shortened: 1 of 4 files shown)


python/tests/generic_alias_1.py
```diff
@@ -1,18 +1,22 @@
...
+from typing import Literal, TypedDict, TypeVar, Generic
+from typing_extensions import TypeAliasType
 
 from typechat import python_type_to_typescript_schema
 
-class First[T](TypedDict):
+T = TypeVar("T", covariant=True)
+
+
+class First(Generic[T], TypedDict):
```
DanielRosenwasser, line 9: Currently TypeChat only understands generics using 3.12 syntax. It doesn't consider types that extend `Generic[...]` or generic protocols as adding type parameters. So I think we either find a way to make some of these tests only run in 3.12, or temporarily disable them.

pcdeadeasy (approved):
Looks good.

python/tests/generic_alias_1.py
hillary-mutisya: I updated the TypeChat code to read type parameters for the classes that extend Generic[...] from the __parameters__ field. With this, the generic_alias_1.py test now outputs type info correctly:
```typescript

// Entry point is: 'FirstOrSecond'
// TypeScript Schema:

type FirstOrSecond<T> = First<T> | Second<T>

interface Second<T> {
    kind: "second";
    "second_attr": T;
}

interface First<T> {
    kind: "first";
    "first_attr": T;
}
```

hillary-mutisya (commented):
Updated based on PR feedback

DanielRosenwasser (approved):

DanielRosenwasser (approved):

PR #184, opened 2024-02-20
Create a top-level `devcontainer.json`.

DanielRosenwasser:
Not at all a devcontainer expert, but I'm trying to create a dev environment that should work well between our Node.js and Python workloads.

@hillary-mutisya let me know if this is similar to what you put together. The only concern that I have is that the creation time is higher than I would expect (it seems like this might still be building Python even with the Oryx flag?)

(CC @samruddhikhandale @eljog)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,7 +1,7 @@
 // For format details, see https://aka.ms/devcontainer.json. For config options, see the
 // README at: https://github.com/devcontainers/templates/tree/main/src/ubuntu
 {
-    "name": "Ubuntu",
+    "name": "TypeChat Development (Python and TypeScript)",
     // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
     "image": "mcr.microsoft.com/devcontainers/python:1-3.12",
 
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,35 +3,27 @@
 {
 	"name": "Ubuntu",
 	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
-	"image": "mcr.microsoft.com/devcontainers/base:jammy",
+	"image": "mcr.microsoft.com/devcontainers/python:1-3.12",
+
+	// Features to add to the dev container. More info: https://containers.dev/features.
 	"features": {
 		"ghcr.io/devcontainers/features/node:1": {
 			"nodeGypDependencies": true,
 			"version": "lts",
 			"nvmVersion": "latest"
 		},
-		"ghcr.io/devcontainers/features/python:1": {
-			"installTools": true,
-			"version": "3.12.2",
-			"useOryxIfAvailable": "false"
-		},
-		"ghcr.io/devcontainers-contrib/features/hatch:2": {}
+		"ghcr.io/devcontainers-contrib/features/hatch:2": {
+			"version": "latest"
+		}
 	},
 
-	// Features to add to the dev container. More info: https://containers.dev/features.
-	// "features": {},
-
 	// Use 'forwardPorts' to make a list of ports inside the container available locally.
-	// "forwardPorts": [],
-
-	// Use 'postCreateCommand' to run commands after the container is created.
-	// "postCreateCommand": "uname -a",
+	"forwardPorts": [7860, 7861, 7862, 7863, 7864, 7865, 7866, 7867, 7868, 7869, 7870],
 
 	// Configure tool-specific properties.
 	"customizations": {
 		"vscode": {
 			"extensions": [
-				"ms-python.python",
 				"ms-python.black-formatter"
 			],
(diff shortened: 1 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,47 @@
...
+	// "forwardPorts": [],
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	// "postCreateCommand": "uname -a",
+
+	// Configure tool-specific properties.
+	"customizations": {
+		"vscode": {
+			"extensions": [
+				"ms-python.python",
```
samruddhikhandale, line 34: You can remove this extension, the Python Feature already adds it. See https://github.com/devcontainers/features/tree/main/src/python#vs-code-extensions

.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for more information:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+# https://containers.dev/guide/dependabot
+
+version: 2
+updates:
+ - package-ecosystem: "devcontainers"
```
samruddhikhandale, line 9: 🎉

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,47 @@
...
+	"image": "mcr.microsoft.com/devcontainers/base:jammy",
+	"features": {
+		"ghcr.io/devcontainers/features/node:1": {
+			"nodeGypDependencies": true,
+			"version": "lts",
+			"nvmVersion": "latest"
+		},
+		"ghcr.io/devcontainers/features/python:1": {
+			"installTools": true,
+			"version": "3.12.2",
```
samruddhikhandale, line 15:
```suggestion
			"version": "3.12",
```

Same here, as you are not using `universal`, feel free to pin it to `3.12` (you no longer need to pin to a specific patch version)

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,47 @@
...
+			"useOryxIfAvailable": "false"
+		},
+		"ghcr.io/devcontainers-contrib/features/hatch:2": {}
+	},
+
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	// "features": {},
```
samruddhikhandale, line 22:
```suggestion
```

Features object is already added

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,47 @@
...
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [],
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	// "postCreateCommand": "uname -a",
+
```
samruddhikhandale, line 29:
```suggestion
```

Same here!

samruddhikhandale (commented):
So happy to see a dev container contribution to this repo  🎉
Thank you!!

.devcontainer/devcontainer.json
hillary-mutisya: Should we be using 3.11 for the container? That is the minimum version for typechat.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,47 @@
...
+			"useOryxIfAvailable": "false"
+		},
+		"ghcr.io/devcontainers-contrib/features/hatch:2": {}
+	},
+
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	// "features": {},
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [],
```
hillary-mutisya, line 25: Please add the forwarded ports from the python container: 
```
 "forwardPorts": [7860,7861,7862,7863,7864,7865,7866,7867,7868,7869,7870],
```
These are required for the notebook integrations

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,47 @@
...
+				"ms-python.python",
+				"ms-python.black-formatter"
+			],
+			"settings": {
+				"python.defaultInterpreterPath": "/workspaces/TypeChat/python/.hatch/typechat/bin/python3.12"
+			}
+		}
+	},
+
+    "postCreateCommand": "cd python; hatch config set dirs.env.virtual .hatch; hatch config update; hatch env create"
```
hillary-mutisya, line 43: Does this work without 
```
"onCreateCommand": "pip install hatch",
``
Unless the container already has hatch, we would need to install it before invoking it.

.devcontainer/devcontainer.json
DanielRosenwasser: It's the minimum, but I will work on making the workspace 3.12-friendly so that we can make sure 3.12 tests run correctly.

hillary-mutisya (approved):

PR #189, opened 2024-02-23
Add missing step in readme

hillary-mutisya:
- Fix error in the path to the typescript edition of typechat
- Add step to npm install before npm build


DanielRosenwasser (approved):
D'oh!

PR #186, opened 2024-02-21
Create pytest compatible tests for existing tests. 

pcdeadeasy:
- [x] Create test_* files that invoke the classes from other modules and run the python to typescript schema generator.
- [x] TBD, add a way to do snapshot testing
- [ ] Invoke the tests as part of the CI/CD pipeline

You can run the tests a couple of ways:
```
cd python/tests
pytest
```

or 

```
cd python
hatch run test
```

Issue #165, opened 2024-01-05
Implement some snapshot testing for our Python schema translation


DanielRosenwasser:
Currently we have a few tests that can be run manually, but we have no testing infrastructure to make sure things don't regress.

I would advocate for something like snapshot/baseline testing which we use within TypeScript.

What we do on TypeScript is have outputs always checked against a "reference" folder which is treated like the source of truth. Newer outputs are emitted to a "local"/"current" folder which can be compared to the reference folder.

When generated files are different from what's in the reference folder, the test runner fails. These folders can be diffed easily using a tool like Beyond Compare or WinMerge. When changes are considered desirable, we have have a way of accepting the baselines using a command (or they can be manually copied over).

I'm not as familiar with testing in the Python ecosystem. [syrupy](https://pypi.org/project/syrupy/) and [pytest-snapshot](https://pypi.org/project/pytest-snapshot) seem to be a plugins for pytest which had a recent release. There's also [turnt](https://github.com/cucapra/turnt) which is all about capturing compiler outputs and the like.

DanielRosenwasser:
https://github.com/microsoft/TypeChat/pull/186 and https://github.com/microsoft/TypeChat/pull/209 have sufficiently implemented snapshotting and CI.

PR #192, opened 2024-02-25
Use `OrderedDict` in place of `dict` and `set` when order matters.

DanielRosenwasser:
This should address the issues in https://github.com/microsoft/TypeChat/pull/186#discussion_r1501340232 so that our schema translation output is deterministic, which is especially important for snapshots and baselines (see #165).


pcdeadeasy (approved):
Looks good, thanks for the change!

PR #193, opened 2024-02-27
Add support for Python tuple types

DanielRosenwasser:
This change adds support for translating tuple types in Python.

A fixed-length tuple is translated to a corresponding array-backed tuple in a TypeScript schema.

```py
target_type = [int, str] # is `[number, string]` in TS

target_type = [int] # is `[number]` in TS

target_type = [()] # is `[]`, the empty tuple, in TS
```

An unbounded homogeneous tuple is simply translated into an array.

```py
target_type = [int, ...] # is 'number[]' in TS
```

Any unsupported form is typically translated into `any[]`.


pcdeadeasy (approved):
Looks good!

PR #197, opened 2024-02-28
Fix health data example.

pcdeadeasy:
Package.json was pointing to the older typechat version. The module did not build with the newer version.

### DIFF
--- a/typescript/examples/healthData/package.json
+++ b/typescript/examples/healthData/package.json
@@ -12,7 +12,7 @@
   "license": "MIT",
   "dependencies": {
     "dotenv": "^16.3.1",
-    "typechat": "^0.0.10"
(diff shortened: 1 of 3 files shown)


hillary-mutisya (approved):
Looks good

DanielRosenwasser (approved):
I'm not a fan of the reformatting (or Prettier's recommendations in general) - I'd prefer if this was just a version update.

Otherwise, it looks good.

pcdeadeasy:
> I'm not a fan of the reformatting (or Prettier's recommendations in general) - I'd prefer if this was just a version update.
> 
> Otherwise, it looks good.

It’s more than a version update, few changes. Prettifier was on in my dev environment. Will check later today.

PR #198, opened 2024-02-29
Replace usage of OpenAI SDK with direct API calls via httpx.

DanielRosenwasser:
This change aligns the behavior of TypeChat for Python with that of its TypeScript counterpart with respect to Azure OpenAI usage.

While TypeChat for TypeScript simply accepted the fully-formed endpoint URL for chat completions [as documented here](https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#chat-completions), the `openai` package made it so that individual components of that URL had to be threaded through piecemeal (i.e. the deployment name, the model name, the version number, and the API key). What that meant was (at least as far as I could tell), the logic didn't work for Azure OpenAI.

This PR makes it so that TypeChat only cares about the API key and the fully formed URL being placed in your `.env` file:

```
AZURE_OPENAI_API_KEY={your-api-key}
AZURE_OPENAI_ENDPOINT=https://{your-resource-name}.openai.azure.com/openai/deployments/{deployment-id}/chat/completions?api-version={api-version}
```

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -23,6 +23,7 @@ classifiers = [
 ]
 dependencies = [
   "pydantic>=2.5.2",
+  "httpx>=0.27.0",
 ]
 
 [project.urls]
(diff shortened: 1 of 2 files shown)


pcdeadeasy (approved):
Looks good

PR #205, opened 2024-03-02
Fix `typescript/healthData` and search upward for `.env` files in TS examples.

DanielRosenwasser:
This PR does two things.

First, it fixes the `typescript/healthData` example by bumping the version and fixing up the example. In principle, I expect the changes should be the same as #197, but without any formatting changes other than import ordering

Second, it changes all our examples to use combine `dotenv` with `find-config` [as suggested here](https://github.com/motdotla/dotenv/issues/238#issuecomment-348017037) so that `.env` files can be placed at the top level of the repo.

### DIFF
--- a/typescript/examples/calendar/package.json
+++ b/typescript/examples/calendar/package.json
@@ -12,9 +12,11 @@
   "license": "MIT",
   "dependencies": {
     "dotenv": "^16.3.1",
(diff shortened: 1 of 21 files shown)


pcdeadeasy (approved):
Looks good

PR #211, opened 2024-03-14
Specify `override` for every TS dotenv call.

DanielRosenwasser:
Fixes issues if (for some reason) environment variables are already specified.

### DIFF
--- a/typescript/examples/calendar/src/main.ts
+++ b/typescript/examples/calendar/src/main.ts
@@ -9,7 +9,7 @@ import { CalendarActions } from './calendarActionsSchema';
 
 const dotEnvPath = findConfig(".env");
 assert(dotEnvPath, ".env file not found!");
-dotenv.config({ path: dotEnvPath });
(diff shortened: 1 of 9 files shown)


DanielRosenwasser:
@jakebailey mentions this may not be desirable since users should be able to declare an environment variable before running the program. On the other hand, I wonder how many people do that with a different key?

jakebailey:
My shell is configured (I think a default in zsh, or maybe just oh-my-zsh?) to not store any command that starts with a space into history, so I can run command with secrets in my terminal without putting them anywhere and that's pretty useful.

PowerShell can do it to with this in `profile.ps1`:

```pwsh
Set-PSReadLineOption -AddToHistoryHandler {
    param($command)
    if ($command -like ' *') {
        return $false
    }
    return $true
} 
```

Of course, these are just examples. Not really sure how we are expecting them to be run.

DanielRosenwasser:
I needed this because of https://github.com/microsoft/vscode-python/issues/23078

PR #225, opened 2024-03-29
Update README.md

tychotic:
Fix the change directory (cd) step in the install directions to correctly  go into the typescript directory before npm install. Note that there is a similar readme in the parent directory that was already correct.


tychotic:
This doc bug actually blocked me for a day (not a heavy npm user) until I stumbled across the corrected readme in the parent directory. I'm guessing others will follow the docs in the examples directory like I did and get a build error.

DanielRosenwasser:
Sorry about that, but thank you!

Issue #215, opened 2024-03-17
[python] Why the hatch dependency?


pamelafox:
Will that be always needed? Or is that just for the examples? I've personally never used it nor seen it as a requirement. If its for venv management reasons, we typically use venv since its in standard lib.

DanielRosenwasser:
So I'm not an expert in all the individual tools, but I was looking for something that would manage

* dependencies
* dev-time dependencies
* virtual environments that could deal with both
* editable installs for writing samples/tests that imported the library
* easy packaging/publishing

Since I've gotten my hands dirtier trying to make things work, I've learned enough to do a decent amount of this with just `pip` and `venv`, but solutions like Poetry, Hatch, and Rye made the onramp easier, and still manage a decent amount of it.

Hatch was recommended to me, I think in part because it's being developed by PyPA and tries to align with their recommendations; however, it still feels like it's early days even if the project is promising. I am open to switching to Poetry given the UX, docs, and familiarity that it has, but also open to something more minimal.

pamelafox:
Yes, I recommend showing the pip/venv route, since that is likely to be what developers are expecting to see (and they could adapt for poetry/hatch/etc). I think uisng hatch probably makes particular sense for *contributing* to TypeChat for the dev-time deps, editable install, packaging, but I don't think it makes as much sense as the default path for a *user* of TypeChat.

For example: I got `hatch shell` working in my terminal, but then when I moved to VS Code, I had no idea where the venv got made so I couldn't get VS Code to use the hatch-made venv, and running hatch shell didnt seem to work. I could have read the hatch docs to try to figure it out, but it would have been more straightforward if I had setup via simple venv.

PR #228, opened 2024-04-04
Make it easier to use purely `pip` and `venv` instead of requiring Hatch

DanielRosenwasser:
In #207, I made it so that hatch would create a `.venv` in the repository root. The reason was to make common scenarios work better for editors (most which currently don't understand hatch).

Still, lots of people don't use Hatch, and in #215, @pamelafox pointed out that it's nice to have control over this stuff. So this PR makes it possible to use use pip directly with our `pyproject.toml`. To do so, dev-time dependencies for TypeChat, along with dependencies for our examples, are moved into their own respective arrays in `optional-dependencies`.

Since pip allows you to install "extras" (see [here](https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/#install-extras), [here](https://packaging.python.org/en/latest/specifications/dependency-specifiers/#extras), and [here](https://pip.pypa.io/en/stable/cli/pip_install/#:~:text=Install%20a%20package%20with%20extras)), developers can just write something like the following:

```sh
cd TypeChat/python
python -v venv ../.venv
source ../.venv/bin/activate
pip install .[dev,examples]
```

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -24,37 +24,18 @@ Install TypeChat:
 pip install typechat
 ```
 
-You can also develop TypeChat from source, which needs [Python >=3.11](https://www.python.org/downloads/).
-
-It can be managed through [hatch](https://hatch.pypa.io/1.6/install/):
+You can also develop TypeChat from source, which needs [Python >=3.11](https://www.python.org/downloads/), [hatch](https://hatch.pypa.io/1.6/install/), and [Node.js >=20](https://nodejs.org/en/download):
 
 ```sh
 git clone https://github.com/microsoft/TypeChat
 cd TypeChat/python
 hatch shell
-```
-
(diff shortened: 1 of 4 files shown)


python/README.md
```diff
@@ -16,18 +16,46 @@ After defining your types, TypeChat takes care of the rest by:
...
 ```sh
 git clone https://github.com/microsoft/TypeChat
 cd TypeChat/python
 hatch shell
 ```
 
+Or with a custom virtual environment:
+
+```sh
+git clone https://github.com/microsoft/TypeChat
```
pamelafox, line 40: This is the stuff I would typically put in CONTRIBUTING.md since most developers won't need it, once this is on Pypi. Your call though!

PR #235, opened 2024-04-17
Use readline for interactive input

gvanrossum:
Importing the readline module and using input() instead of sys.stdin turns on command line editing, including history.
For a REPL dinosaur like myself this is essential. :-)


DanielRosenwasser (approved):

DanielRosenwasser:
Thank you!

PR #237, opened 2024-04-17
Add Python usage docs

DanielRosenwasser:


site/src/docs/python/basic-usage.md
```diff
@@ -0,0 +1,246 @@
...
+        Returns a `Success[T]` object containing the object if validation was successful.
+        Otherwise, returns a `Failure` object with a `message` property describing the error.
+        """
+        ...
+```
+
+To construct a validator, we just have to pass in the type we defined:
+
+```py
+import schema as sentiment
```
gvanrossum, line 174: This naming convention feels a bit confusing. I'd say *either* we recommend putting the schema definition in a file `schema.py` with a toplevel class named after the application, and then the driver code can just write `from schema import Sentiment`. *Or* we recommend using the application name for the schema file, e.g. `sentiment.py`. But using `schema.py` for the filename and then importing as the app name feels just a tad convoluted for a Python developer.

site/src/docs/python/basic-usage.md
```diff
@@ -0,0 +1,246 @@
...
+translator = TypeChatJsonTranslator(model, validator, sentiment.Sentiment)
+```
+
+When we are ready to translate a user request, we can call the `translate` method.
+
+```ts
+translator.translate("Hello world! 🙂");
+```
+
+We'll come back to this.
```
gvanrossum, line 195: Yeah, this is the expository style I like better! :-)

site/src/docs/python/basic-usage.md
```diff
@@ -0,0 +1,246 @@
...
+    ...
+
+file_path = sys.argv[1] if len(sys.argv) == 2 else None
+await process_requests("😀> ", file_path, request_handler)
+```
+
+`process_requests` takes 3 things.
+First, there's the prompt prefix - this is what a user will see before their own text in interactive scenarios.
+You can make this playful.
+We like to use emoji here. 😄
```
gvanrossum, line 213: Why? I still haven't figured out how to *type* emoji on a keyboard -- I only know how to do it on my phone. :-(

site/src/docs/python/basic-usage.md
```diff
@@ -0,0 +1,246 @@
...
+
+When we are ready to translate a user request, we can call the `translate` method.
+
+```ts
+translator.translate("Hello world! 🙂");
+```
+
+We'll come back to this.
+
+## Creating the Prompt
```
gvanrossum, line 197: "Prompt" is an ambiguous term here. Does it refer to the user input prompt (only used when file_path is None) or the prompt for the LLM?

site/src/docs/python/basic-usage.md
```diff
@@ -0,0 +1,246 @@
...
+
+TypeChat exports a `process_requests` function that makes it easy to experiment with TypeChat.
+Depending on its second argument, it either creates an interactive command line prompt (if given `None`), or reads lines in from the given a file path.
+
+```ts
+async def request_handler(message: str):
+    ...
+
+file_path = sys.argv[1] if len(sys.argv) == 2 else None
+await process_requests("😀> ", file_path, request_handler)
```
gvanrossum, line 207: In Python it's more common to name such a variable or argument `filename` or `file`.

site/src/docs/python/basic-usage.md
```diff
@@ -0,0 +1,246 @@
...
+    result = await translator.translate(message)
+    if isinstance(result, Failure):
+        print(result.message)
+    else:
+        result = result.value
+        print(f"The sentiment is {result.sentiment}")
+```
+
+We're calling the `translate` method on each string and getting a response.
+If something goes wrong, TypeChat will retry requests up to a maximum specified by `retry_max_attempts` on our `model`.
```
gvanrossum, line 239: This is a feature of the Model class, right? May be helpful to mention that.

site/src/docs/python/basic-usage.md
```diff
@@ -0,0 +1,246 @@
...
+import schema as sentiment
+validator = TypeChatValidator(sentiment.Sentiment)
+```
+
+## Creating a JSON Translator
+
+A `TypeChatJsonTranslator` brings all these concepts together.
+A translator takes a language model, a validator, and our expected type, and provides a way to translate some user input into objects following our schema.
+To do so, it crafts a prompt based on the schema, reaches out to the model, parses out JSON data, and attempts validation.
+Optionally, it will craft repair prompts and retry if validation fails.
```
gvanrossum, line 183: I would love to read more about the repair process. I expect that in practice one might have to tweak this.

gvanrossum (commented):
Here's a very nitpicky review of the basic Python docs (sorry :-).

site/src/docs/python/basic-usage.md
gvanrossum: Eh, let's keep it. The key is to explain where `schema` comes from rather than presenting code that will fail if you copy it into a .py file and run it.

site/src/docs/python/basic-usage.md
```diff
@@ -0,0 +1,246 @@
...
+TypeChat is currently a small library, so we can get a solid understanding just by going through the following example:
+
+```py
+import asyncio
+
+import sys
+import schema as sentiment
+from typechat import Failure, TypeChatJsonTranslator, TypeChatValidator, create_language_model, process_requests
+
+async def main():
```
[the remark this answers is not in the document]
gvanrossum: Scratch that. The rest of the example needs to be indented. And a call to asyncio.run(main()) added to the end.

site/src/docs/python/basic-usage.md
```diff
@@ -0,0 +1,246 @@
...
+env_vals = dotenv_values()
```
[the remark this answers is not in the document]
gvanrossum: The example should just have `from dotenv import dotenv_values`.

gvanrossum (commented):
I am going to fix all these and more in a new commit.

site/src/docs/python/basic-usage.md
```diff
@@ -0,0 +1,246 @@
...
+        ...
+```
+
+then you should be able to try TypeChat out with such a model.
+
+The key thing here is providing a `complete` method.
+`complete` is just a function that takes a `string` and eventually returns a `string` if all goes well.
+
+For convenience, TypeChat provides two functions out of the box to connect to the OpenAI API and Azure's OpenAI Services.
+You can call these directly.
```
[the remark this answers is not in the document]
gvanrossum: Uh, let's keep this structure, it's similar to the TS docs, and I don't feel like fixing this and making sure there are no further forward references.

site/src/docs/python/basic-usage.md
```diff
@@ -0,0 +1,246 @@
...
+```
+
+For even more convenience, TypeChat also provides a function to infer whether you're using OpenAI or Azure OpenAI.
+
+```ts
+def create_language_model(vals: dict[str, str | None]) -> TypeChatLanguageModel:
+```
+
+With `create_language_model`, you can populate your environment variables and pass them in.
+Based on whether `OPENAI_API_KEY` or `AZURE_OPENAI_API_KEY` is set, you'll get a model of the appropriate type.
```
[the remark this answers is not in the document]
gvanrossum: Skip that.

site/src/docs/python/basic-usage.md
gvanrossum: Later.

robgruen (approved):

PR #239, opened 2024-04-18
Fix/trailing commas from chatgpt

faceleg:
ChatGPT sometimes returns JSON with trailing spaces, which breaks the parser. The repair attempts do not take this into account.

I've copied in the strip trailing comma function from here: https://github.com/nokazn/strip-json-trailing-commas/blob/main/src/index.ts (MIT) and added a test to prove it works:

- https://github.com/microsoft/TypeChat/blob/7c6837444ded6c3ebf88020050d2d0434c049c51/typescript/src/ts/strip-whitespace.ts
- https://github.com/faceleg/TypeChat/blob/7ff8b444d3003d33466d298d3b6865a79374aaf0/typescript/src/ts/strip-whitespace.test.ts

**Trailing commas are stripped here:** 
https://github.com/microsoft/TypeChat/blob/7c6837444ded6c3ebf88020050d2d0434c049c51/typescript/src/typechat.ts#L140

**Example broken response:**
```json
{
  "items": [
    {
      "id": 1,
      "text": "驳回",
      "exampleSentences": [
        "法官驳回了他的上诉请求。",
        "公司决定驳回他的辞职申请。",
        "政府部门驳回了他的建议。",
      ],
      "partsOfSpeech": "verb"
    },
    {
      "id": 2,
      "text": "驳回",
      "exampleSentences": [
        "他对这个提案的驳回感到失望。",
        "这个决定的驳回引起了公众的不满。",
        "他的建议被驳回了，让他感到沮丧。",
      ],
      "partsOfSpeech": "noun"
    }
  ]
}
```
**Example prompt that generated this response:**

> You are a helpful vocabulary learning assistant who helps users generate example sentences in Mandarin for language learning. You understand that in Mandarin, words can serve different parts of speech depending on context.
> 
> Please find the possible usages this word: 休想, and generate 3 example sentences for each usage.
> 
> The sentences should be medium or longer length and complexity of HSK5 or higher. Each sentence must contain the the word. All sentences provided for the word must be unique.
> 
> JSON must be returned as an array of objects, with one object per part of speech for the word. You must return valid JSON. The array of sentences must not have a trailing comma.

This is the project I'm using TypeChat on: https://github.com/faceleg/ankiai, forked from https://github.com/mhujer/ankiai.

### DIFF
--- a/typescript/jest.config.js
+++ b/typescript/jest.config.js
@@ -1,0 +1,5 @@
+module.exports = {
+    preset: 'ts-jest',
+    testEnvironment: 'node',
+    testMatch: ['**/*.test.ts'],
+};
--- a/typescript/package-lock.json
+++ b/typescript/package-lock.json
@@ -13,7 +13,10 @@
         "./examples/*"
       ],
       "devDependencies": {
-        "@types/node": "^20.10.4"
+        "@types/jest": "^29.5.12",
+        "@types/node": "^20.10.4",
+        "jest": "^29.7.0",
+        "ts-jest": "^29.1.2"
       },
       "engines": {
         "node": ">=18"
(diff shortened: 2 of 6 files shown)


faceleg:

robgruen:
I'm wary of adding arbitrary regex parsing to the LLM generated text.  What if my payload is invalidly formatted JSON (by design), this will, without hesitation modify that text as well potentially breaking my use-case.  A more correct way to fix it would be in JSON parsing

LLMs have also gotten much better since this initial PR was submitted, I wonder if this is still occurs as much? Also, there's some merge conflicts. Therefore I will close it out and if you'd like to pursue it please submit a new PR. I wonder if we could prompt engineer this rather than taking indiscriminate action on the text of the response?

PR #238, opened 2024-04-18
New (Python) example: simple drawings (rectangle, ellipse, arrow)

gvanrossum:
For me, this works best with openai model gpt-3.5-turbo; gpt-4 seems worse (?!).

<details>
<summary>Here's a sample session (no longer valid without history support)</summary>

```
PATH=/opt/homebrew/opt/ccache/libexec:/opt/homebrew/bin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/Library/Apple/usr/bin:/usr/local/share/dotnet:~/.dotnet/tools:/Users/guido/.vscode/extensions/ms-python.python-2024.4.1/python_files/deactivate/bash:/Users/guido/TypeChat/python/.venv/bin:/Library/Frameworks/Python.framework/Versions/3.13/bin:/Library/Frameworks/Python.framework/Versions/3.11/bin:/Library/Frameworks/Python.framework/Versions/3.12/bin:/Users/guido/.nvm/versions/node/v19.8.1/bin:/opt/homebrew/opt/ccache/libexec:/opt/homebrew/bin

The default interactive shell is now zsh.
To update your account to use zsh, please run `chsh -s /bin/zsh`.
For more details, please visit https://support.apple.com/kb/HT208050.
(.venv) ~/TypeChat/python$ PYTHONPATH=src python examples/drawing/demo.py 
~> draw three squares of side 50 in a diagonal
--------- NEXT REQUEST ---------
[...]

PR #240, opened 2024-04-20
Actually test the Python translator

DanielRosenwasser:
This PR adds a test for Python translators by mocking a model and snapshotting the log.

Based on feedback from @gvanrossum where he pointed out that we don't actually use the repaired prompt (I broke this in #203). Some of the fixes are already in https://github.com/microsoft/TypeChat/pull/238.

### DIFF
--- a/python/tests/__snapshots__/test_translator.ambr
+++ b/python/tests/__snapshots__/test_translator.ambr
@@ -1,0 +1,84 @@
+# serializer version: 1
+# name: test_translator_with_immediate_pass
+  list([
+    dict({
+      'kind': 'CLIENT REQUEST',
+      'payload': '''
+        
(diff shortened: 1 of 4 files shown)


python/tests/__snapshots__/test_translator.ambr
```diff
@@ -0,0 +1,137 @@
+# serializer version: 1
```
gvanrossum, line 1: What is this file? I don't see it mentioned anywhere, but maybe you have some clever test setup that uses this?

DanielRosenwasser: Yeah, [syrupy](https://github.com/tophat/syrupy) is a pytest plugin that makes it easy-ish to generate snapshot/baseline tests. They have a default snapshot format called Amber (which I've never heard of)

Issue #39, opened 2023-07-21
Conversations rather than one-offs


claus-topholt-private:
Any thoughts on how to use TypeChat in conversation-style interactions? In my use case, there is a need to go back and forth with the LLM, refining queries. In your coffee shop example, something like this:

User: Two tall lattes. The first one with no foam.
Assistant: Two tall lattes coming up.
User: The second one with whole milk. Actually make the first one a grande.
Assistant: One grande latte, one tall latte with whole milk. Coming up.

steveluc:
Great question.  We are experimenting with adding an conversation history section to the beginning of the prompt that would provide an array of user and assistant entries.  A complication is that in TypeChat the LLM output is a formal representation of user intent.  Because of this, it's not the right thing to put in the Assistant part of the history.  Instead, an abstraction or summary of the application's output to the user should be in the Assistant part of the history.  How to do this varies by application.  For example the music application may output a relatively long track list in response to a search request.  Rather than consuming tokens by placing in the track list verbatim, we are looking at putting in a summary of the track list or just the information that a track list with k entries was printed.  We need to gain some more experience with this aspect of conversation history, but we will be doing this one relatively soon.

claus-topholt-private:
Wonderful - that's what I've been working towards outside of TypeChat, but you guys save me a ton of time!

coderfengyun:
> Any thoughts on how to use TypeChat in conversation-style interactions? In my use case, there is a need to go back and forth with the LLM, refining queries. In your coffee shop example, something like this:
> 
> User: Two tall lattes. The first one with no foam. Assistant: Two tall lattes coming up. User: The second one with whole milk. Actually make the first one a grande. Assistant: One grande latte, one tall latte with whole milk. Coming up.

This example would be more realistic with some modifications. For example, the bot could be asked to summarize the user's needs and seek confirmation from the user.

coderfengyun:
[...]

PR #241, opened 2024-04-22
Proposed alternate prompting scheme

gvanrossum:
This is my proposal to address #39. It also fixes the issue I flagged in #240 about the user input being repeated twice. I tried it with my demo from #238 and that seems to work.

The prompt structure is as follows:

- System: "You are a ..., up to and including the schema, with instructions for how to format the JSON"
- prompt_prelude (user and bot messages)
- User: (the user request)

I don't understand how your test tools work (I'm a dinosaur, I don't know how hatch or toml or pipenv work, I can handle pytest though).

@DanielRosenwasser

### DIFF
--- a/python/src/typechat/_internal/translator.py
+++ b/python/src/typechat/_internal/translator.py
@@ -64,14 +64,14 @@ class TypeChatJsonTranslator(Generic[T]):
 
         messages: list[PromptSection] = []
 
-        messages.append({"role": "user", "content": input})
+        messages.append({"role": "system", "content": self._create_request_prompt(input)})
+
         if prompt_preamble:
             if isinstance(prompt_preamble, str):
                 prompt_preamble = [{"role": "user", "content": prompt_preamble}]
-            else:
-                messages.extend(prompt_preamble)
+            messages.extend(prompt_preamble)
 
-        messages.append({"role": "user", "content": self._create_request_prompt(input)})
+        messages.append({"role": "user", "content": input})
 
         num_repairs_attempted = 0
         while True:
@@ -103,11 +103,7 @@ You are a service that translates user requests into JSON objects of type "{self
 ```
 {self._schema_str}
 ```
-The following is a user request:
-'''
-{intent}
-'''
-The following is the user request translated into a JSON object with 2 spaces of indentation and no properties with the value undefined:
+You translate each user request into a JSON object with 2 spaces of indentation and no properties with the value undefined.
 """
         return prompt
 


DanielRosenwasser:
So this matches my intuition around how to put all of this together. While we have no way of testing for a regression, I'd be okay with running with something like this for now. The thing I'll note is that when @ahejlsberg tried something like this with the system prompt, he got slightly worse results.

That said, he guessed part of this was probably tied to the schema falling further and further behind in history. So maybe something to alleviate this would be to instead do:

- prompt_prelude (user and bot messages)
- System: "You are a ..., up to and including the schema, with instructions for how to format the JSON"
- User: (the user request)

Basically, move the system prompt immediately before the user request. 
_____

> I don't understand how your test tools work

No worries. To most quickly fix the build:

```sh
cd TypeChat/python
pip install -e .[dev,examples]
pytest --snapshot-update
git add tests
git commit -m "Update snapshots."
```

DanielRosenwasser:
It probably also makes sense to make the repair prompt a `system` message as well.

DanielRosenwasser:
One of the things I tried in https://github.com/microsoft/TypeChat/tree/systemPromptPython (ea0e0d8a3a4e2d5ea1982f35ae8cc730468eec73) was to take the approach here after fixing some issues. As @ahejlsberg mentioned, OpenAI unfortunately doesn't do a great job when you switch to the system prompts. [Trying things out on the coffee shop example input file](https://github.com/microsoft/TypeChat/blob/ea0e0d8a3a4e2d5ea1982f35ae8cc730468eec73/python/examples/coffeeShop/input.txt), it generally got more "wrong".

One idea we could try is to expose the (very basic) logic we use to parse out JSON so that people can override translate and do whatever they want.

But otherwise, the only other approaches I could advise to implement something like https://github.com/microsoft/TypeChat/pull/238 is to experiment with preambles to talk about past state.

gvanrossum:
Thanks for the feedback!

## Preamble issues

My problem with the preamble is that it is *prefixed* to the prompt containing the schema, which makes the conversation a bit awkward, as it goes like this (ignoring system messages):

- user: _draw three squares_
- bot: {some JSON}
- user: _make them bigger_
- bot: {some JSON}
- user: Please translate user requests into JSON using this schema: {schema JSON}. Here is a user question: _make them red_

What I want is this (because my intuition tells me that it's a more natural dialog):

- user: Please translate user requests into JSON using this schema: {schema JSON}. Here is a user question: _draw three squares_
- bot: {some JSON}
- user: _make them bigger_
- bot: {some JSON}
- user: _make them red_


## About system messages:

My (rudimentary) understanding of system messages is that it's meant to set a behavior or personality that's hard(er) to subvert by users trying to be clever -- e.g. the system prompt would tell the model not to be racist. In TypeChat's case, I think that at best the prompt that sets the schema should be a system prompt, but the repair prompts should probably just be user prompts.

The [docs](https://platform.openai.com/docs/guides/text-generation/chat-completions-api) say a little bit about this. It makes me think that it's not worth experimenting too much with this in particular.

I also just noticed "[JSON Mode](https://platform.openai.com/docs/guides/text-generation/json-mode)" in the docs, which may or may not be useful for TypeChat.

ahejlsberg:
@gvanrossum FWIW, a while back I did a multi-turn experiment where I didn't include previous replies in the prompt, but rather just repeated the request with the additional user input concatenated to the the previous user input (separated by newlines). So, your example would have a single `user` prompt section with text like:

```
Please translate user requests into JSON using this schema:
{schema JSON}.
Here is a user question:
_draw three squares_
_make them bigger_
_make them red_
```

It seemed to work quite well. Wonder if that would work in your case?

gvanrossum:
I tried concatenating the user inputs, and it works okay — I’m not completely happy, but I wasn’t with my old approach either.

I’d think that it would not work well with ChatGPT’s caching scheme; though I think that TypeChat makes that hard anyway by not giving the user the exact output from the model.

It makes me wonder what the prompt_preamble is intended for, though.

________________________________

python/src/typechat/_internal/translator.py
gvanrossum, line 100: The 'intent' argument is no longer used. Since this is an internal function, can't you remove it?

python/src/typechat/_internal/translator.py
```diff
@@ -64,14 +64,14 @@ async def translate(self, input: str, *, prompt_preamble: str | list[PromptSecti
...
         messages: list[PromptSection] = []
 
-        messages.append({"role": "user", "content": input})
+        messages.append({"role": "system", "content": self._create_request_prompt(input)})
+
         if prompt_preamble:
             if isinstance(prompt_preamble, str):
                 prompt_preamble = [{"role": "user", "content": prompt_preamble}]
-            else:
-                messages.extend(prompt_preamble)
```
gvanrossum, line 72: Do I understand correctly that the original code had a bug, where if `prompt_preamble` was a string, it would be ignored? Was that bug reported in any issue?

gvanrossum (commented):
Shouldn't there be a test showing something has improved? Or are there no tests using live models?

I have to say I'm not sure about the AI-generated report -- I feel overwhelmed by all the data, and I don't know what it's trying to look for or what's significant. Also, why is it testing the non-Python implementations? Any changes there would be just noise, since this PR doesn't affect them.

python/src/typechat/_internal/translator.py
robgruen: This was your change from a while ago so I can't speak to that.  I think you mean to have the "intent" to be a separate user message after the JSON translation string which you're proposing as the system prompt.

robgruen: here are no live tests at the moment, that's on the TODO list.  The AI generated report I sent you compares the existing prompting scheme in comparison with the one proposed here.  Since both the C# and TypeScript implementations have the same prompting scheme I wanted to ensure that we see the same results across implementations.

The net takeaway is that the net token count goes down by a few percentage points as does the overall retry count.  With the current set of models this approach is marginally better given the sample schemas we have in the repo at this time.  I would like to flush it out a bit more before declaring victory and making a final determination. I favor this approach since there is clear separation between the messages whereas the existing schema is slightly more coupled.

PR #254, opened 2024-07-26
Enable passing images along with text for models that support multimodal input

hillary-mutisya:
GPT-4-vision, GPT-4-omni and GPT-4-turbo allow multi-modal input, where images and text can be passed in the prompt. To support this, the content section of the prompt has an array of objects instead of just a string.

### DIFF
--- a/typescript/src/model.ts
+++ b/typescript/src/model.ts
@@ -36,7 +36,16 @@ export type ImagePromptContent = {
 };
 
 export type ImageUrl = {
+    /*
+     * This could be a URL to a hosted image, or the base64-encoded image content.
+     */
     url: string;
+    
+    /*
+     * Controls how the model processes the image and generates its textual understanding.
+     * In "low res" mode, the model treats the image as 512x512px, while "high res" mode considers
+     * the image at full resolution.
+     */
     detail?: "auto" | "low" | "high";
 };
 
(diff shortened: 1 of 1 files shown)


typescript/src/model.ts
```diff
@@ -135,7 +168,11 @@ function createFetchLanguageModel(url: string, headers: object, defaultParams: o
             const response = await fetch(url, options);
             if (response.ok) {
                 const json = await response.json() as { choices: { message: PromptSection }[] };
-                return success(json.choices[0].message.content ?? "");
+                if (typeof json.choices[0].message.content  === 'string') {
```
DanielRosenwasser, line 171:
```suggestion
                if (typeof json.choices[0].message.content === 'string') {
```

typescript/src/model.ts
```diff
@@ -135,7 +168,11 @@ function createFetchLanguageModel(url: string, headers: object, defaultParams: o
...
+                    return success(json.choices[0].message.content ?? "");
+                } else {
+                    return error(`REST API unexpected response format :  ${JSON.stringify(json.choices[0].message.content)}`);
```
ahejlsberg, line 174: Change to `` `REST API unexpected response format: ${JSON.stringify(json.choices[0].message.content)}` ``

ahejlsberg: That is, get rid of the extra blanks.

DanielRosenwasser:
```suggestion
                    return error(`REST API unexpected response format: ${JSON.stringify(json.choices[0].message.content)}`);
```

hillary-mutisya: Fixed

PR #260, opened 2024-08-30
Add Car Assistant Natural Language example

brclarkdev:
A set of two examples focused on leveraging typechat to improve the performance of in vehicle AI assistants.


brclarkdev:
@microsoft-github-policy-service agree company="Microsoft"

robgruen:
While we appreciate the contribution these examples are more heavy than we'd like and seem like they will be evolving regularly.  I recommend creating a new OSS project and placing these examples in that repo.

PR #265, opened 2024-09-13
docs: update example readme

yugasun:


gvanrossum (approved):
LGTM. The old instructions seem to date from before there were multiple supported languages, when the only language was TS and it lived in the toplevel directory.

gvanrossum (approved):

PR #272, opened 2025-04-24
add getJsonSchema() to TypeChatJsonValidator

mmoskal:
This is a little experiment, to possibly allow using structured outputs with TypeChat.

Note that you would want to only use this as a grammar for constraining the output, not pass the schema to the model - the TypeScript file is probably much better for this, as the schema is 3x bigger than TypeScript source (with indentation; without indentation schema is 1.5x bigger). OpenAI-compatible endpoints will often pass the schema to the model using chat template.

This works for all example schemas (excluding Zod, but it can be handled natively I guess) and all schemas in TypeAgent.


robgruen:
@copilot , can you see if there's any gaps in the tstoJS schema conversion.

PR #277, opened 2025-10-21
Change OpenAI API endpoints from Completions to Responses for TypeScript

fedtti:
This PR simply move from the Completions API to the newer Responses API endpoint (cf. https://platform.openai.com/docs/guides/migrate-to-responses). They should be fully retro-compatible due to the messages' nature (cf. https://platform.openai.com/docs/guides/migrate-to-responses?update-item-definitions=responses&update-multiturn=responses#migrating-from-chat-completions).


robgruen:
subsumed by #305

PR #338, opened 2026-06-01
chore(dependabot): add npm/pip ecosystems and group updates

TalZaccai:
## Problem

The existing `.github/dependabot.yml` only declared the `devcontainers` and `github-actions` ecosystems. All 9 currently-open security alerts in this repo target `typescript/package-lock.json` (`tar`, `qs`, `minimatch`, `@tootallnate/once`) — a manifest under an ecosystem Dependabot wasn't watching — so **no update PRs have ever been opened for any of them**. The Python tree (`/python/pyproject.toml`) is similarly uncovered.

## Change

* Add `npm` coverage for `/typescript`, `/typescript/examples/*`, and `/site`
* Add `pip` coverage for `/python`
* Add grouping rules across all ecosystems so multiple updates in a given run collapse into a single PR. Routine version updates and security-advisory-triggered updates are grouped **separately** (via `applies-to: security-updates`) so security PRs can be prioritised on their own.
* Major-version bumps continue to open one PR per package so reviewers can evaluate them individually.

## Expected outcome

After merge, the next Dependabot run (weekly Monday) should open at most one grouped npm-security PR covering the existing 9 alerts.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,11 +6,61 @@
 
 version: 2
 updates:
+  # npm — typescript library + examples + docs site.
+  # All security alerts in this repo to date have been against
+  # typescript/package-lock.json, which previously had no ecosystem entry
+  # here, so Dependabot never opened any update PRs for them.
+  - package-ecosystem: "npm"
+    directories:
+      - "/typescript"
+      - "/typescript/examples/*"
+      - "/site"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    labels:
+      - "dependencies"
+    groups:
+      npm-production:
+        dependency-type: "production"
+        update-types: ["minor", "patch"]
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -6,11 +6,69 @@
 
 version: 2
 updates:
+  # npm — typescript library + examples + docs site.
```
robgruen, line 12: comment out of date

robgruen (approved):

.github/dependabot.yml
TalZaccai: Good catch — fixed in the latest commit. Replaced the auto-generated "To get started..." boilerplate with a real header that documents the routine / security / major-bump grouping policy this file actually applies.

PR #430, opened 2026-08-27
test: call the configured API when available

### LINKED ISSUE #328: Update tests to call actual API if available, mock otherwise

mikemikimike:
## Summary

Closes #328.

The model integration test suite now performs a real language-model request when `OPENAI_API_KEY` and `OPENAI_MODEL` are configured. In environments without those variables, the existing deterministic mocked tests remain the default, so local and CI runs do not require credentials.

## Tests

- `npm run build`
- `npx tsc -p test`
- `node --test out/validate.test.js out/zod.test.js out/program.test.js tests/model.test.mjs`
- Result: 113 passed, 1 skipped (no API credentials configured)


robgruen:
Thanks for picking up #328!  Can you please address copilot's comment, it seems valid to me.

Your test results show 113 passed / 1 skipped, which means this path hasn't been exercised yet — could you re-run with credentials set and paste the output? 

Second, gating purely on OPENAI_API_KEY being present will make npm test silently issue billable calls for anyone who's run the examples. Please add a separate explicit opt-in (e.g. TYPECHAT_LIVE_TESTS=1), bound the timeout and retry count, and move the test into its own describe so it never interacts with the fetch mocking."

mikemikimike:
Addressed the review feedback in commit f6b4a6defd1bd1e90621b39119435789843ac205.

- Preserved the native global fetch at module initialization and restore it in teardown instead of deleting it, so later live tests can use the runtime implementation.
- Moved the network test into its own `createLanguageModel live API integration` suite.
- Added an explicit `TYPECHAT_LIVE_TESTS=1` opt-in, so normal test runs never make billable calls accidentally.
- Bounded the live test with a 30-second timeout and zero retries.

Validation:
- `npm test` with `TYPECHAT_LIVE_TESTS=1`, `OPENAI_MODEL=qwen3.5-4b`, and `OPENAI_ENDPOINT=http://127.0.0.1:1234/v1/chat/completions`
- LM Studio served the local qwen3.5-4b model through its OpenAI-compatible API.
- Result: 114 passed, 0 skipped, 0 failed.

The live request was verified against the local LM Studio OpenAI-compatible endpoint rather than the hosted OpenAI API; no OpenAI credential was available in this environment.

